PR #1, opened 2020-06-19
Update Readme

mikaelkrief:
Remove the Installation Section:
Because the provider is integrated in the registry, there is no need to install it manually

## All Submissions:

* [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x ] I have updated the documentation accordingly.

Update the reademe by removing the Installation section


nmiodice:
Hoping to fix the lint issues in #11

PR #17, opened 2020-06-26
Add build definition auth

### LINKED ISSUE #7: [New Functionality] Allow build definition authorization to repository
### Description

Currently the resource `azuredevops_resource_authorization` only supports service endpoints.

Feature request for adding build definition authorization to a repository.

If you are extending a build pipeline from one repository to another, ADO will prompt you to permit access the first time it is run

* azuredevops_resource_authorization

azuredevops_resource_authorization

```hcl
resource "azuredevops_resource_authorization" "pipeline_auth" {
  project_id  = azuredevops_project.project.id
  resource_id = azuredevops_build_definition.plan.id
  authorized  = true
}
```

### References

* [resource_authorization](https://www.terraform.io/docs/providers/azuredevops/r/resource_authorization.html)

image

### Community Note

### LINKED ISSUE #68: Missing "path" parameter documentation in docs
### Community Note

### Description

Missing Missing `path` parameter documentation in docs in https://www.terraform.io/docs/providers/azuredevops/r/build_definition.html

### Affected Resource(s)

* `azuredevops_build_definition`

EliiseS:
## All Submissions:

Add support to authorize resources for a build

Issue Numbers: 
- #7
- Fixes #68  

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+    "files.eol": "\n",
```
tmeckel: Go files should always use Unix line endings but other files should not, in case someone what's to code on Windows. Line endings for `.go`files are already handled by the top level `.gitattributes`. https://github.com/terraform-providers/terraform-provider-azuredevops/blob/257c1c7e63f4be92bba1182e81e5f45242fd4902/.gitattributes#L4

tmeckel (commented):
I've fixed the linting errors in my PR #18 already

.vscode/settings.json
```diff
@@ -1,6 +1,39 @@
 {
-    "go.testTags": "all",
-    "gopls.env": {
-        "GOFLAGS": "-tags=all"
```
tmeckel, line 4: If you remove the `gopls.env`, you'll lose intellisense for unit and acceptance tests on VSCode

EliiseS: REALLY?!?! OH MY GOD, THATS AMAZING! I've been looking for how to fix that issue for AGES. It's been driving me nuts!

I've moved all the settings from the dev container and I didn't notice that this one was gopls one. 

Love it and added back under the other gopls settings. Having more than one setting of the same name just takes the latest.

.vscode/settings.json
EliiseS: Hmm! Good point! I don't know nearly enough about line ending. I'll move this setting back to the dev container, since that's always linux. 

On the point of file endings, shouldn't we add to gitattributes to always use a specific file ending for certain file types? I know that git gets really confused whenever you clone into a different system and you end up with hundreds of files changed.

.vscode/settings.json
tmeckel: Hell, never saw someone freaking out by getting intellisense support 😁 👍

.vscode/settings.json
tmeckel: The infamous, often hated story of different line endings 🥴 😁 Okay, where to start. 🤔 Normally Git is very good in handling different line endings on different operating systems (okay ... different line endings between Windows and the rest 😁 ) automatically. Internally Git stores all files as Unix line endings, but during checkout it converts it to the default line endings of the current operating system. (https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings) By this, even people who use `notepad`, could edit code (text files) that are under control of Git.

But sometimes you have to pin the line endings for a specific file or file type (like `*.go`) because for other (technical) reasons. When you're using the Go language you should be a aware that the people behind Go denied the support for the CRLF line endings on Windows for all tools that come with Go like `gofmt`. That's why I filed a PR (https://github.com/terraform-providers/terraform-provider-azuredevops/commit/406481301ffe5cf10043bafe7001307b49b0c5a3) to pin the line endings to LF for *.go files, so that I could develop using Windows and I don't always run into the issue that Git is complaining about diverted line endings when I reformatted the code with `gofmt` or because the Go AddIn for VSCode was so nice to do that while saving a `*.go` file.

But to me it makes little sense to pin line endings in `.gitattributes` or in the VSCode settings because Git is handling the conversion very well and, as said, ensures that people can use their favorite text editor without any hassle.

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -72,41 +78,62 @@ func resourceResourceAuthorizationRead(d *schema.ResourceData, m interface{}) er
...
+			})
+
+			if err != nil {
+				return err
+			}
+
+			// the authorization does no longer exist
+			if len(*resourceRefs) == 0 {
+				d.SetId("")
+				return nil
```
tmeckel, line 99: Better to have a log message with at least [INFO] level (I'd even rather prefer [WARN] here) so that people like @xuzhang3, who use the TF_LOG extensively, get a clue what happened.

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -72,41 +78,62 @@ func resourceResourceAuthorizationRead(d *schema.ResourceData, m interface{}) er
...
+			}
+
+			flattenAuthorizedResource(d, &(*resourceRefs)[0], projectID, definitionID)
+			return nil
```
tmeckel: I'm friend of sparingly use the `return` statement and have an additional exit point because it's harder to understand a routine if, reading it at the bottom, you're unaware of the possibility that it returned somewhere above. But, and here comes the catch, this heavily depends on the code, because you drive it to the top and always ensure to have only one single exit point at the end of every function, you could end up with code like that 😁 

```c
function()
{
    HRESULT error = S_OK;

    if(SUCCEEDED(Operation1()))
    {
        if(SUCCEEDED(Operation2()))
        {
            if(SUCCEEDED(Operation3()))
            {
                if(SUCCEEDED(Operation4()))
                {
                }
                else
                {
                    error = OPERATION4FAILED;
                }
            }
            else
            {
                error = OPERATION3FAILED;
            }
        }
        else
        {
            error = OPERATION2FAILED;
        }
    }
    else
    {
        error = OPERATION1FAILED;
    }

    return error;
}
```

But here it would make sense to introduce two additional `else` blocks and to remove two `return` statements

image

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -115,54 +142,63 @@ func resourceResourceAuthorizationDelete(d *schema.ResourceData, m interface{})
...
 	}
 
-	return &resourceRef, d.Get("project_id").(string), nil
+	return &resourceRef, d.Get("project_id").(string), d.Get("definition_id").(int), nil
 }
 
-func sendAuthorizedResourceToAPI(clients *client.AggregatedClient, resourceRef *build.DefinitionResourceReference, project string) error {
+func sendAuthorizedResourceToAPI(clients *client.AggregatedClient, resourceRef *build.DefinitionResourceReference, projectID string, definitionID int) error {
 	ctx := context.Background()
+	if definitionID == 0 {
```
tmeckel, line 193: Introduce an else and remove the return inside the `if` block

tmeckel (changes requested):

azuredevops/internal/service/build/resource_resource_authorization.go
EliiseS: Good idea :)

azuredevops/internal/service/build/resource_resource_authorization.go
EliiseS: Done. Feel free to complain about the log message content :D

.vscode/settings.json
EliiseS: I think it handles them less well, when you clone on windows and move to files to linux (devcontainer), no?

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -120,55 +154,64 @@ func resourceResourceAuthorizationDelete(d *schema.ResourceData, m interface{})
...
 		return fmt.Errorf(msgErrorFailedResourceUpdate, err)
 	}
 
 	return resourceResourceAuthorizationRead(d, m)
 }
 
-func flattenAuthorizedResource(d *schema.ResourceData, authorizedResource *build.DefinitionResourceReference, projectID string) {
+func flattenAuthorizedResource(d *schema.ResourceData, authorizedResource *build.DefinitionResourceReference, projectID string, definitionID int) {
 	d.SetId(*authorizedResource.Id)
 	d.Set("resource_id", *authorizedResource.Id)
```
tmeckel: Can you remove all pointer dereferences when `d.Set` is called, because the `Set` function is very well capable of handling `nil`pointers. This will also minimize the possibility of a `panic` when the API might change an return an empty `nil`value for a computed or optional field.

https://github.com/hashicorp/terraform-plugin-sdk/blob/621b20845686c999865f87682650d38f191b5483/helper/schema/resource_data.go#L164

azuredevops/internal/service/build/resource_resource_authorization_test.go
```diff
@@ -39,103 +40,146 @@ var resourceReferenceNotAuthorized = build.DefinitionResourceReference{
...
-		Resources: &[]build.DefinitionResourceReference{resourceReferenceAuthorized},
-		Project:   &projectId,
-	}
-	buildClient.
-		EXPECT().
-		AuthorizeProjectResources(clients.Ctx, expectedArgs).
-		Return(nil, errors.New(expectedMessage)).
-		Times(1)
-	return r, resourceData, clients
+var tests = []struct {
```
tmeckel, line 64: This is slick!!! Like it!! 🚀 ❤️

azuredevops/internal/acceptancetests/resource_resource_authorization_test.go
```diff
@@ -15,26 +15,65 @@ func TestAccResourceAuthorization_CRUD(t *testing.T) {
...
 
-	resourcesHCL := testutils.HclServiceEndpointGitHubResource(projectName, serviceEndpointName)
+	serviceEndpointHCL := testutils.HclServiceEndpointGitHubResource(projectName, serviceEndpointName)
 	authedHCL := testutils.HclResourceAuthorization("azuredevops_serviceendpoint_github.serviceendpoint.id", true)
 	unAuthedHCL := testutils.HclResourceAuthorization("azuredevops_serviceendpoint_github.serviceendpoint.id", false)
 
+	tfAuthNode := "azuredevops_resource_authorization.auth"
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testutils.PreCheck(t, &[]string{"AZDO_GITHUB_SERVICE_CONNECTION_PAT"})
```
tmeckel, line 25: I'm always thinking that we might have a check at the start of an AccTest, which skips the test if a required environment variable isn't set, instead of failing the test at all. The trade off could be that we don't detect misconfigured pipelines. The advantage would be that contributors could run AccTests more easily.

tmeckel (changes requested):

azuredevops/internal/acceptancetests/resource_resource_authorization_test.go
EliiseS: I was thinking that too! But, I guess if we had a way to detech we were in a pipeline and then not skip them?

tmeckel: Pipelines runners typically add some kind of environment variables to each job. 
[GitLab](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html) has `CI`
[Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) has `CI`
[Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml)  (Microsoft) is as always doing something else 😁 
GitHub had some discussions about it https://github.com/actions/runner/issues/368

But because people mainly use the `CI` environment variable, and you're always able to add environment variables to a  job or runners definition, I'd test for `CI=[true|1|yes|on]` (case insensitive regex of course 😀) to check if the code runs (or it's pretended) in a pipeline.

azuredevops/internal/service/build/resource_resource_authorization.go
EliiseS: Done for this, not doing everything, you maniac :D

azuredevops/internal/acceptancetests/resource_resource_authorization_test.go
EliiseS: Added ;)

tmeckel: :grinning: cool, but I thought it should be implemented in the way that if CI is not set **and not** the required environment variable of the test, **than** it's skipped.

That ensures that  if we're running in a pipeline the test will fail if the environment variable of the test is not set but a developer can set the environment variable of the test and run the acceptance test outside a pipeline. :smiley:

EliiseS: I don't follow?

tmeckel: Okay  sorry ... used my mobile to answer back on your last comment and I had the impression you did something different than changing `testutils.PreCheck`

I thought about an additional function like so 

```go
// SkipIfNotEnabled will skip a test if one of the environment variables are not set and the test is not running inside a pipelie (environment variable CI != "")
func SkipIfNotEnabled(t *testing.T, additionalEnvVars *[]string) {
	if os.Getenv("CI") == "" && additionalEnvVars != nil {
		for _, variable := range *additionalEnvVars {
			if strings.EqualFold(os.Getenv(variable), "") {
				t.Skipf("Skipping test %s, because environment variable is `%s` not set and required to run this test.", t.Name(), variable)
			}
		}
	}
}
```

And having a call to this function at the first line of an acceptance test

```go
func TestAccResourceAuthorization_CRUD(t *testing.T) {
	testutils.SkipIfNotEnabled(t, &[]string{"AZDO_GITHUB_SERVICE_CONNECTION_PAT"})

        ...
}
```

I'm a bit uncomfortable with your approach because firstly, it might skip **all** acceptance tests, including core tests that only require the Azure DevOps REST API (AZDO_PERSONAL_ACCESS_TOKEN) and need no external credentials or similar. 

And secondly, I am not a fan of extending (changing) the semantics of an existing function. Your modification to 'PreCheck' now causes tests to be skipped and the function did not do that before.

EliiseS: Honestly, I don't think adding a new function call to all tests is a good solution (a lot of tests). Especially when you also have to reenter the env vars required.

I agree it should fail on basic on core tests. I can change it to fail on core tests or I can remove it completely.

tmeckel: Maybe we should just bury the idea again. It also somehow collides with the Go build tags, which you can use, at least via the powershell script `acctest.ps1` with parameter `-tags`, to skip certain acceptance tests. Not as convenient as the idea here, but also easy to use.

EliiseS: So, I've gone and made the testing optional. I've placed it under a new make command and updated the docs

.vscode/launch.json
```diff
@@ -5,16 +5,22 @@
...
+                "-test.run",
+                "^${selectedText}$"
+            ],
             "env": {
-                "TF_ACC": "1"
+                "TF_ACC": "1",
             },
-            "args": [],
-            "buildFlags": "-v -tags=all"
+            "buildFlags": "-v -tags=all",
```
tmeckel, line 21: I'd recommend to add `-gcflags="all=-N -l"` as a build flags so that debug code is generated. Otherwise sometimes `dlv` (or another debugger) isn't able to display variable contents, because the Go compiler optimized away the needed information by the debugger. I added this once to the `build.ps1` and `build.sh` files already.

https://golang.org/doc/gdb#Introduction

README.md
```diff
@@ -109,17 +109,45 @@ You can also cross-compile if necessary:
...
+  ```sh
+  make testacc
+  ```
+
+* Run a subset using a prefix
+
+  ```sh
+  make testacc TESTARGS='-run=TestAccBuildDefinitionBitbucket_Create'
+  ```
+
```
tmeckel: Perhaps we should introduce a variable in the makefile so that people can limit the executed acceptance tests via tags like so:

`make testacc TESTTAGS='core data_agent_pool exclude_resource_group'`

You can achieve this already by executing acceptance tests with the `acctests.ps1` script using the `-Tags` parameter but I think because of consistency reasons we should support this in the makefile as well.

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -33,17 +34,23 @@ func ResourceResourceAuthorization() *schema.Resource {
...
-				Optional:     true,
+				Required:     true,
 				Description:  "id of the resource",
 				ValidateFunc: validation.NoZeroValues,
 			},
+			"definition_id": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Description:  "id of the build definition",
+				ValidateFunc: validation.NoZeroValues,
```
tmeckel: `validate.NoEmptyStrings` 🙏 😁

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -73,20 +80,39 @@ func resourceResourceAuthorizationRead(d *schema.ResourceData, m interface{}) er
...
+				Type:    authorizedResource.Type,
+				Id:      authorizedResource.Id,
+			})
+
+			if err != nil {
+				return err
+			}
+
+			if len(*resourceRefs) == 0 {
+				log.Printf("[WARN] The authorization with ID '%s' no longer exists. Setting Id to empty \n", *(authorizedResource.Id))
```
tmeckel: Can you add this message to line 120 as well?

image

tmeckel (changes requested):

azuredevops/internal/service/build/resource_resource_authorization.go
tmeckel: Just checked if the Terraform Plugin SDK provides something similar, so that we can remove code from our own code base. Not surprisingly there is `validation.StringIsNotWhiteSpace` 😁

EliiseS: Oh my god! Gaaaaaaaaaaaaaaaah

.vscode/launch.json
EliiseS: but, I can see the variable contents when I debug? Can you test out it's not working as intended?

tmeckel: Well personally I had the issue with variables that contain pointers to structs as I remember correctly. So perhaps with simple variables its not an issue at all.

image

GNUmakefile
```diff
@@ -44,8 +44,13 @@ test: fmtcheck
     		xargs -t -n4 go test -tags "all" $(TESTARGS) -timeout=30s -parallel=4
 
 testacc: fmtcheck
+	@echo "==> Sourcing .env file if avaliable"
+	if [ -f .env ]; then set -o allexport; source .env; set +o allexport; fi
 	TF_ACC=1 go test -tags "all" $(TEST) -v $(TESTARGS) -timeout 120m
 
+testacc-skip-tests: fmtcheck
+	TF_ACC=1 SKIP_NOT_SET_TESTS=1 go test -tags "all" $(TEST) -v $(TESTARGS) -timeout 120m
```
xuzhang3: Why add `SKIP_NOT_SET_TESTS`? If this environment variable not set,  raise an panic if enough for users.

GNUmakefile
```diff
@@ -44,8 +44,13 @@ test: fmtcheck
     		xargs -t -n4 go test -tags "all" $(TESTARGS) -timeout=30s -parallel=4
 
 testacc: fmtcheck
+	@echo "==> Sourcing .env file if avaliable"
+	if [ -f .env ]; then set -o allexport; source .env; set +o allexport; fi
```
xuzhang3: Is this `.env` file bind with vscode? If it is, don't recommend add this config file here.  `make` should not  tied to any specific specify development tool configuration.

README.md
```diff
@@ -109,17 +109,45 @@ You can also cross-compile if necessary:
...
 
-The majority of tests in the provider are Acceptance Tests - which provisions real resources in Azure. It's possible to run the entire acceptance test suite by running `make testacc` - however it's likely you'll want to run a subset, which you can do using a prefix, by running:
+With VSCode Golang extension you can also run and debug the tests using `run test`, `debug test` `run package tests`, `run file tests` buttons.
 
-```sh
-make testacc SERVICE='resource' TESTARGS='-run=TestAccAzureRMResourceGroup' TESTTIMEOUT='60m'
-```
+#### Acceptance tests
+
+The majority of tests in the provider are acceptance tests - which provisions real resources in Azure Devops and Azure. To run any acceptance tests you need to set `AZDO_ORG_SERVICE_URL`, `AZDO_PERSONAL_ACCESS_TOKEN` environment variables, some test have additional environment variables required to run. You can find out the required environment variables by running the test.
```
xuzhang3: In my perspective,  developers who want contribute to this provider,  mostly they focus on the resource they interested.  `make testacc` will run all acceptance tests, this will get developers confused with the error messages.

azuredevops/internal/acceptancetests/testutils/commons.go
```diff
@@ -51,8 +51,12 @@ func PreCheck(t *testing.T, additionalEnvVars *[]string) {
 	}
 
 	for _, variable := range requiredEnvVars {
-		if strings.EqualFold(os.Getenv(variable), "") {
-			t.Fatalf("`%s` must be set for acceptance tests!", variable)
+		if _, ok := os.LookupEnv(variable); !ok {
+			// Skip tests with missing additional environment variables when SKIP_NOT_SET_TESTS set
+			if _, ok := os.LookupEnv("SKIP_NOT_SET_TESTS"); ok && additionalEnvVars != nil {
```
xuzhang3: @EliiseS  Can you provide your background concerns here. I cannot get the point why add `SKIP_NOT_SET_TESTS`.
For example: we set `AZDO_PERSONAL_ACCESS_TOKEN`, `AZDO_ORG_SERVICE_URL` and `SKIP_NOT_SET_TESTS`. If we run `make testacc`, all service connections and other  acceptance tests relies on environment configuration will be ignored  because they need the environment configures (constraint: configurations are not set in .tf file).

website/docs/r/build_definition.html.markdown
```diff
@@ -78,6 +78,7 @@ The following arguments are supported:
 
 - `project_id` - (Required) The project ID or project name.
 - `name` - (Optional) The name of the build definition.
+- `path` - (Optional) The path of the build definition.
```
xuzhang3:
```suggestion
- `path` - (Optional) The folder path of the build definition.
```

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -120,55 +154,64 @@ func resourceResourceAuthorizationDelete(d *schema.ResourceData, m interface{})
...
-func sendAuthorizedResourceToAPI(clients *client.AggregatedClient, resourceRef *build.DefinitionResourceReference, project string) error {
+func sendAuthorizedResourceToAPI(clients *client.AggregatedClient, resourceRef *build.DefinitionResourceReference, projectID string, definitionID int) error {
 	ctx := context.Background()
-
-	_, err := clients.BuildClient.AuthorizeProjectResources(ctx, build.AuthorizeProjectResourcesArgs{
-		Resources: &[]build.DefinitionResourceReference{*resourceRef},
-		Project:   &project,
-	})
+	var err error
+	if definitionID == 0 {
```
xuzhang3, line 203: If `definition_id` not configured. `nil` may be a more proper way. Constraint is stronger than `0`

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -120,55 +154,64 @@ func resourceResourceAuthorizationDelete(d *schema.ResourceData, m interface{})
...
-	d.Set("authorized", *authorizedResource.Authorized)
+	d.Set("resource_id", authorizedResource.Id)
+	d.Set("type", authorizedResource.Type)
+	d.Set("authorized", authorizedResource.Authorized)
 	d.Set("project_id", projectID)
+	d.Set("definition_id", definitionID)
 }
 
-func expandAuthorizedResource(d *schema.ResourceData) (*build.DefinitionResourceReference, string, error) {
+func expandAuthorizedResource(d *schema.ResourceData) (*build.DefinitionResourceReference, string, int, error) {
```
xuzhang3: return param `error` can be removed, always `nil`

azuredevops/internal/service/build/resource_resource_authorization.go
```diff
@@ -120,55 +154,64 @@ func resourceResourceAuthorizationDelete(d *schema.ResourceData, m interface{})
 	// because the resource to delete might have had this parameter set to true, we overwrite it
 	authorizedResource.Authorized = converter.Bool(false)
 
-	err = sendAuthorizedResourceToAPI(clients, authorizedResource, projectID)
+	err = sendAuthorizedResourceToAPI(clients, authorizedResource, projectID, definitionID)
```
xuzhang3: `expandAuthorizedResource` will always return error as`nil`. `if` statement is meaningless here

azuredevops/internal/service/build/resource_resource_authorization.go
xuzhang3, line 193: If `definition_id ` not configured. `nil ` may be a more proper way, `definition_id ` is not configured, should do this operation.

xuzhang3 (changes requested):
@EliiseS  This PR is great.  I left some comments, :D

azuredevops/internal/acceptancetests/testutils/commons.go
tmeckel: As you say correctly, contributors only focus on the one resource they want to contribute. If a developer now wants to run acceptance tests locally, it would be a great simplification if the tests for which no infrastructure/token/account is available (e.g. Bitbucket) could simply be disabled. 

We already have a solution here using the build tags in the `_test.go` files and the PowerShell script `acctest.ps1` but maybe we need a more simple solution. I think the easier it is for the provider to develop, the more likely it is that there will be many people who want to do it. The entry hurdles must not be too high.

azuredevops/internal/service/build/resource_resource_authorization.go
tmeckel: No, the function returns a potential error
image

azuredevops/internal/service/build/resource_resource_authorization.go
tmeckel: Issue here is that if you do a `d.Get("definition_id").(int)` you'll get the `zero` value. The other alternative, as you mentioned, would be using `d.GetOk`, check if `true` return an `*int` or `false` than return `nil` for an `*int`. Because 0 is an invalid value for a definition ID in DevOps, I can live with the implementation here. 

image

azuredevops/internal/service/build/resource_resource_authorization.go
EliiseS: That's a type int, though soo, I think that's valid?

EliiseS: I'll change the one above it, that's a string though

README.md
EliiseS: Like that?

GNUmakefile
```diff
@@ -45,11 +46,11 @@ test: fmtcheck
...
 testacc: fmtcheck
 	@echo "==> Sourcing .env file if avaliable"
-	if [ -f .env ]; then set -o allexport; source .env; set +o allexport; fi
-	TF_ACC=1 go test -tags "all" $(TEST) -v $(TESTARGS) -timeout 120m
+	if [ -f .env ]; then set -o allexport; . ./.env; set +o allexport; fi; \
+	TF_ACC=1 go test -tags "$(TESTTAGS)" $(TEST) -v $(TESTARGS) -timeout 120m
 
-testacc-skip-tests: fmtcheck
-	TF_ACC=1 SKIP_NOT_SET_TESTS=1 go test -tags "all" $(TEST) -v $(TESTARGS) -timeout 120m
+testacc-skip-tests: export SKIP_NOT_SET_TESTS=1
```
tmeckel: Like this tactic to re-use a Makefile target ❤️ 👍

README.md
tmeckel: Of course! This is great! 👍 🚀

azuredevops/internal/service/build/resource_resource_authorization.go
tmeckel: Ha ... You paid attention! But with `NoZeroValue` you can set a negative value. Perhaps it's better to use `IntAtLeast(1)` or something.

tmeckel (changes requested):

azuredevops/internal/service/build/resource_resource_authorization.go
xuzhang3: function `expandAuthorizedResource` not `sendAuthorizedResourceToAPI` , at line 160. I comment on the wrong code line :D

azuredevops/internal/acceptancetests/testutils/commons.go
xuzhang3: we can run the specify acceptance test with 
`make testacc TESTARGS='-run=TestAccServiceEndpointBitBucket' `   or
`TF_ACC=1 go test -tags "all" ./... -run=TestAccServiceEndpointBitBucket -v  -timeout 120m`
`-run` param will match test case with function name prefix.

README.md
```diff
@@ -109,17 +109,45 @@ You can also cross-compile if necessary:
...
+
+* Run all tests, but skip tests that have missing additional environment variables
+
+  ```sh
+  make testacc-skip-tests
+  ```
+
+* With VSCode Golang extension you can also run the tests using `run test`, `run package tests`, `run file tests` buttons above the test
+
+##### Debugging acceptant tests
```
xuzhang3: Can we move this to `doc/debugging.md` or a new document like `development.md`? This section is about how to debug with vscode. But we have users use different develop tools `Goland`,`vim`, `emacs` etc.

azuredevops/internal/acceptancetests/testutils/commons.go
tmeckel: @xuzhang3  I agree, especially because @EliiseS added with 77a7add4442fba55a55178f3a0e40dd6642da8c9 the possibility to specify build tags for the `GNUMakefile` (as you already can with the `acctests.ps1` script) to limit (exclude) the number of acceptance tests being run. So I think we can then do without `SKIP_NOT_SET_TESTS`.

EliiseS: I added this because I'd like to run all tests locally, and skip 'specialty' tests I haven't set up yet, such as ones requiring 	`AZDO_DOCKERREGISTRY_SERVICE_CONNECTION_USERNAME`, `AZDO_DOCKERREGISTRY_SERVICE_CONNECTION_EMAIL`, `AZDO_DOCKERREGISTRY_SERVICE_CONNECTION_PASSWORD`, ect.

Basically when I have to run acceptance tests manually locally for my or other peoples PRs.

.vscode/launch.json
EliiseS: Do you still have issues? I was not able to reproduce this:

image

azuredevops/internal/acceptancetests/testutils/commons.go
tmeckel: @EliiseS yes that's really convenient but as I said, after you added 77a7add, you can limit the acceptance test to execute by specifying their build tags (or exclusion tags). So basically after 77a7add there isn't a real need for `SKIP_NOT_SET_TESTS`. I would be fine having both possibilities here but @xuzhang3 is more strict on such things I guess.

README.md
EliiseS: Error message? I'm just updating the existing documentation since it was out of date. I think documenting why the acceptance tests will fail if you run all of them is more than valid. I know I was confused for a while :D 

And documenting that they can be dummy values is more than valid! 😄

.vscode/launch.json
tmeckel: Because I always have the possibility to build a "debug version" of the provider by using the `build[.sh|.ps1]` scripts, or change the `buildFlags` , I don't insist having the `gcflags` inside the VSCode configuration.

EliiseS: Setting the flag results in error:
```sh
go test: gcflags flag may be set only once
run "go help test" or "go help testflag" for more information
Process exiting with code: 1
```

Flag is already set

GNUmakefile
EliiseS: You're right, as I've discovered these values can be set to be dummy values

GNUmakefile
EliiseS: No, it's just a .env file that you can set all of your config in and if it's there will be picked up by the tooling. Not VSCode specific.

.vscode/launch.json
tmeckel: AHA! That's why you never had issues, when you start debugging directly through VSCode and do not by attaching to a running process like I did. The guys at VSCode are clever enough to set this flag to ensure you can see all variables and stuff 👍

azuredevops/internal/acceptancetests/testutils/commons.go
EliiseS: Actually, Xu and I have spoken and he's enlightened me to the fact that almost all of the values used in the acceptance tests can be set as dummy values

tmeckel: But the dummy values will make the tests fail as if the environment variables aren't set. Right?

README.md
EliiseS: Done

README.md
EliiseS: make testacc was previously mentioned as well

.vscode/launch.json
```diff
@@ -5,16 +5,22 @@
...
+            "showLog": true,
+            "envFile": "${workspaceFolder}/.env"
         }
     ]
 }
```
tmeckel: I wanna try to push the number of comments on this PR to be over 100 🤣 🤣  No not really, but ... spotted two files missing a newline at the of each file.

Can you add those settings to `.vscode\settings.json`

```json
"files.trimFinalNewlines": true
"files.insertFinalNewline": true
```

Background: https://thoughtbot.com/blog/no-newline-at-end-of-file

Because people use different editors aside VSCode we should have a (lint-)test which checks if all (text) files have a **single** newline at the end.

.vscode/settings.json
```diff
@@ -1,6 +1,41 @@
...
+        "usePlaceholders": true, // add parameter placeholders when completing a function
+        // Experimental settings
+        "completeUnimported": true, // autocomplete unimported packages
+        "deepCompletion": true // enable deep completion
+    },
+    "cSpell.words": [
+        "hashicorp",
+        "azuredevops"
+    ]
 }
```
tmeckel: Missing newline

azuredevops/internal/acceptancetests/testutils/commons.go
EliiseS: No, all the tests pass. (I know, I was shocked too)

tmeckel (changes requested):

azuredevops/internal/acceptancetests/testutils/commons.go
tmeckel: WTF!?!?!? 😮  What nonsense! 🤷‍♂️ 🤦‍♂️

.vscode/settings.json
EliiseS: gahgsdhdaj

.vscode/launch.json
EliiseS: Gah

.vscode/settings.json
tmeckel: Really!?!?!? Are you serious????? 😮 🤣

GNUmakefile
EliiseS: I'll be merging this PR, because it's gotten a bit out of hand, if this is still an issue I can make a new PR to address this :)

README.md
EliiseS: I'll be merging this PR, because it's gotten a bit out of hand, if this is still an issue I can make a new PR to address this :)

Issue #72, opened 2020-07-13
Github service connection API breaking change


xuzhang3:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Affected Resource(s)

* `azuredevops_serviceendpoint_github`

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

### Expected Behavior

1. create github service connection with  `auth_oauth` success
2. create github service connection with `auth_personal` can verify success from portal after `terrafom apply` .

### Actual Behavior

1. create github service connection with  `auth_oauth` failed
![image](https://user-images.githubusercontent.com/57888764/87289558-63fed000-c52f-11ea-9d17-8fae2362711f.png)
2. create github service  connection with `auth_personal` can create success but failed when verify from portal
![image](https://user-images.githubusercontent.com/57888764/87289655-7f69db00-c52f-11ea-87b3-9b2b33835285.png)

### Steps to Reproduce

1. create github service connection with `auth_oauth`
  - prepare the .tf configure file
  - execute `terraform apply --auto-approve=true`

2. create github service connection with `auth_personal`
  - prepare the .tf configure file
  - execute `terraform apply --auto-approve=true`
  - verify the config from portal

### References

* #0000

PR #74, opened 2020-07-14
GitHub service connection API breaking change

xuzhang3:
## All Submissions:

Issue Number:

#72 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
+++ b/azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
@@ -4,11 +4,62 @@
 package acceptancetests
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/terraform-providers/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
 )
+
+func TestAccServiceEndpointGitHub_CreateWithPersonalToken(t *testing.T) {
+	projectName := testutils.GenerateResourceName()
+	serviceEndpointNameFirst := testutils.GenerateResourceName()
+
+	resourceType := "azuredevops_serviceendpoint_github"
+	tfSvcEpNode := resourceType + ".serviceendpoint"
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, &[]string{"AZDO_GITHUB_SERVICE_CONNECTION_PAT"}) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckServiceEndpointDestroyed(resourceType),
+		Steps: []resource.TestStep{
+			{
+				Config: personTokenConfig(projectName, serviceEndpointNameFirst),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(tfSvcEpNode, "project_id"),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "auth_personal.#", "1"),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "service_endpoint_name", serviceEndpointNameFirst),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "description", "Managed by Terraform"),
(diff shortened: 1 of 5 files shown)


azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
```diff
@@ -4,12 +4,132 @@
...
+					resource.TestCheckResourceAttr(tfSvcEpNode, "service_endpoint_name", serviceEndpointName),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "description", "Managed by Terraform"),
+					testutils.CheckServiceEndpointExistsWithName(tfSvcEpNode, serviceEndpointName),
+				),
+			},
+		},
+	})
+}
+
+func TestAccServiceEndpointGitHub_PersonalTokenUpdate(t *testing.T) {
```
EliiseS, line 39: Doesn't this test cover the `TestAccServiceEndpointGitHub_PersonalTokenBasic`? Do you need to have both?

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
```diff
@@ -4,12 +4,132 @@
...
+					resource.TestCheckResourceAttrSet(tfSvcEpNode, "project_id"),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "auth_oauth.#", "1"),
+					resource.TestCheckResourceAttr(tfSvcEpNode, "service_endpoint_name", serviceEndpointName),
+					testutils.CheckServiceEndpointExistsWithName(tfSvcEpNode, serviceEndpointName),
+				),
+			},
+		},
+	})
+}
+func TestAccServiceEndpointGitHub_OauthUpdate(t *testing.T) {
```
EliiseS, line 97: Isn't the test `TestAccServiceEndpointGitHub_OauthBasic` covered in this test?

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
```diff
@@ -53,3 +173,63 @@ func TestAccServiceEndpointGitHub_CreateAndUpdate(t *testing.T) {
...
+	service_endpoint_name  = "%[1]s"
+	auth_oauth {
+		oauth_configuration_id= "xxxx-xxxx-xxxx-xxxx-xxxx"
+	}
+}`, serviceEndpointName)
+
+	return fmt.Sprintf("%s\n%s", projectResource, serviceEndpointResource)
+}
+
+func oauthConfigUpdate(projectName string, serviceEndpointName string, description string) string {
```
EliiseS, line 222: Do we not want to put them into the hcl.go anymore?

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
xuzhang3: Yes, `basic` is used to create the base service connection then `update` can update the service connection and verify it

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
xuzhang3: I plan to refactor the code
1. The `HCL` test configure is used only in`resource_serviceendpoint_github_test.go`,  no need to make this test data public
2. All test `HCL` in one file, soon we will get a monster file, keep the test data where is belongs will be more clear.

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
xuzhang3: Yes, `basic ` is used to create the base service connection then `update ` will update the service connection just created and verify it

azuredevops/internal/acceptancetests/resource_serviceendpoint_github_test.go
tmeckel: I agree but you still need some place where you put shared HCL definitions. If you add the HCL functions in the `_test.go` files you'll clutter up them as well. Many `_test.go` files are already huge!

PR #89, opened 2020-07-19
Use the default_branch as the name of an initialized branch instead of defaulting to master

jamiemccrindle:
## All Submissions:

Currently if you initialize a repo, it will always create a 'master' branch. If you specify a 'default_branch' that isn't master, setting the default branch will fail. This change means that the value of 'default_branch' is used as the branch name on initialization e.g. if you wanted the branch to be called 'refs/heads/main'. Note: this only changes the behaviour for 'Clean' initialization type. For uninitialized, there won't be a branch to set.

Issue Number: No issue

- [x] No

- [x] No

This does change the behavior in that 'default_branch' now works when initializing repos where previously it didn't. 

n/a

## Other information

No component changes.

### DIFF
--- a/azuredevops/internal/service/git/resource_git_repository.go
+++ b/azuredevops/internal/service/git/resource_git_repository.go
@@ -228,13 +228,17 @@ func createGitRepository(clients *client.AggregatedClient, repoName *string, pro
 }
 
 func initializeGitRepository(clients *client.AggregatedClient, repo *git.GitRepository, defaultBranch *string) error {
+	branchName := converter.ToString(defaultBranch, "")
+	if strings.EqualFold(branchName, "") {
+		branchName = "master"
+	}
 	args := git.CreatePushArgs{
 		RepositoryId: repo.Name,
 		Project:      repo.Project.Name,
 		Push: &git.GitPush{
 			RefUpdates: &[]git.GitRefUpdate{
 				{
-					Name:        converter.String("refs/heads/" + *defaultBranch),
+					Name:        converter.String("refs/heads/" + branchName),
 					OldObjectId: converter.String("0000000000000000000000000000000000000000"),
 				},
 			},
--- a/azuredevops/internal/service/git/resource_git_repository.go
+++ b/azuredevops/internal/service/git/resource_git_repository.go
@@ -230,7 +230,7 @@ func createGitRepository(clients *client.AggregatedClient, repoName *string, pro
 func initializeGitRepository(clients *client.AggregatedClient, repo *git.GitRepository, defaultBranch *string) error {
 	branchName := converter.ToString(defaultBranch, "")
 	if strings.EqualFold(branchName, "") {
-		branchName = "master"
+		branchName = "refs/heads/master"
 	}
 	args := git.CreatePushArgs{
 		RepositoryId: repo.Name,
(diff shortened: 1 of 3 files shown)


xuzhang3:
Hi @jamiemccrindle Thanks for opening this PR.  I try update the repository `default_branch` and I will get an error :
image
Because the branch I try set as default branch does not exist.

jamiemccrindle:
Oh dang. I’ll test further. Can you send me the tf that broke?

xuzhang3:
@jamiemccrindle 
Reproduce steps:
1. `terraform apply`
2. change `default_branch` to another branch  like `refs/heads/release`
3. `terraform apply`

TF config:
```hcl
resource "azuredevops_project" "project" {
  project_name       = "Sample Project"
  visibility         = "private"
  version_control    = "Git"
  work_item_template = "Agile"
}

resource "azuredevops_git_repository" "repo" {
  project_id = azuredevops_project.project.id
  name       = "Sample Empty Git Repository"
  initialization {
    init_type = "Clean"
  }
  default_branch="refs/heads/master"
}
```

jamiemccrindle:
Thank you!

Testing now. I suspect that this may have broken before too i.e. the initalization only happens once and the default branch needs the branch to exist but will confirm.

xuzhang3:
Set `default_branch` attribute `ForceNew: true,` might be a choice for this issue. A new repo will be created and destroy the old one .

jamiemccrindle:
@xuzhang3, have confirmed that this would also break in trunk i.e. changing the default branch after creation to a branch that doesn't exist. If adding `force_new` it useful it might be better to add it in another PR.

EliiseS:
@xuzhang3 @jamiemccrindle Could one of you create an issue for the bug you discovered? 😛

jamiemccrindle:
hi @EliiseS, is it the one for a change of the `default_branch` setting failing on update if the branch doesn't exist? Or an issue for this enhancement?

EliiseS:
The change of the default_branch setting failing on update if the branch doesn't exist @jamiemccrindle  :)

azuredevops/internal/service/git/resource_git_repository.go
```diff
@@ -227,14 +227,18 @@ func createGitRepository(clients *client.AggregatedClient, repoName *string, pro
 	return createdRepository, nil
 }
 
-func initializeGitRepository(clients *client.AggregatedClient, repo *git.GitRepository) error {
+func initializeGitRepository(clients *client.AggregatedClient, repo *git.GitRepository, defaultBranch *string) error {
```
EliiseS, line 230: Could you also create a test for this change?

jamiemccrindle: sure thing! have added a test.

jamiemccrindle:
re: raising an issue for the `default_branch` failing on update if the branch doesn't exist. It's a bit of a tricky one. I think it's reasonable if an update fails if the branch we're setting as the `default_branch` doesn't exist.

what's arguably more interesting is that the `default_branch` works when the repo is first created even if the branch doesn't exist.  this PR fixes it for the case where the initalization type is 'clean' but it's probably still wrong for uninitialized repos. Only problem is that I think that fixing it for uninitialized repos could be considered a breaking change.

xuzhang3:
@jamiemccrindle If you set `initialization` with init_type: Clean`, the  default branch  `master` will be created.  `init_type: uninitialized `   won't create the default branch which means no branch exist for the new created repo.

xuzhang3:
@jamiemccrindle LGTM, thanks for your contribute.

Issue #54, opened 2020-07-03
Terraform crash if initialization not set for git repo


EliiseS:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Crashes with invalid terraform instead of returning an error.

**Valid git_repo:**
```hcl
resource "azuredevops_git_repository" "repo" {
  project_id = azuredevops_project.test.id
  name       = "test-repo"
  initialization {
    init_type = "Clean"
  }
}
```

**Invalid git_repo:**
```hcl
resource "azuredevops_git_repository" "repo" {
  project_id = azuredevops_project.test.id
  name       = "test-repo"
}
```

### Terraform (and Azure DevOps Provider) Version

```sh
10:55 # terraform -v
Terraform v0.12.28
+ provider.azuredevops v0.0.1
```

### Affected Resource(s)

* `azuredevops_git_repository`

### Terraform Configuration Files

```hcl
provider "azuredevops" {
  version = ">= 0.0.1"
}

resource "azuredevops_project" "test" {
  project_name = "Test Project"
}

resource "azuredevops_git_repository" "repo" {
  project_id = azuredevops_project.test.id
  name       = "test-repo"
}

resource "azuredevops_git_repository" "fork-repo" {
  project_id           = azuredevops_project.test.id
  name                 = "forkk-test-repo"
  parent_repository_id = azuredevops_git_repository.repo.id
}
```

### Panic Output
```sh
10:55 # tf apply

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # azuredevops_git_repository.fork-repo will be created
  + resource "azuredevops_git_repository" "fork-repo" {
      + default_branch       = (known after apply)
      + id                   = (known after apply)
      + is_fork              = (known after apply)
      + name                 = "forkk-test-repo"
      + parent_repository_id = (known after apply)
      + project_id           = (known after apply)
      + remote_url           = (known after apply)
      + size                 = (known after apply)
      + ssh_url              = (known after apply)
      + url                  = (known after apply)
      + web_url              = (known after apply)
    }

[...]

PR #92, opened 2020-07-20
Make `initialization` a required field in `azuredevops_git_repository`

beandrad:
Instead of crashing when no initialization type is provided.

## All Submissions:

* [] ~~I have updated the documentation accordingly~~ -> Documentation already implies behavior.

Issue Number: #54

- [ ] Yes
- [x] No

- [ ] Yes
- [X] No

### DIFF
--- a/azuredevops/internal/service/git/resource_git_repository.go
+++ b/azuredevops/internal/service/git/resource_git_repository.go
@@ -97,7 +97,7 @@ func ResourceGitRepository() *schema.Resource {
 			},
 			"initialization": {
 				Type:     schema.TypeSet,
-				Optional: true,
+				Required: true,
 				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -394,7 +394,6 @@ func expandGitRepository(d *schema.ResourceData) (*git.GitRepository, *repoIniti
 
 	var initialization *repoInitializationMeta = nil
 	initData := d.Get("initialization").(*schema.Set).List()
-
 	// Note: If configured, this will be of length 1 based on the schema definition above.
 	if len(initData) == 1 {
 		initValues := initData[0].(map[string]interface{})
--- a/azuredevops/internal/service/git/resource_git_repository_test.go
+++ b/azuredevops/internal/service/git/resource_git_repository_test.go
@@ -111,31 +111,8 @@ func TestGitRepo_FlattenExpand_RoundTrip(t *testing.T) {
 
 	resourceData := schema.TestResourceDataRaw(t, ResourceGitRepository().Schema, nil)
 	resourceData.SetId(gitRepo.Id.String())
-	flattenGitRepository(resourceData, &gitRepo)
-
-	expandedGitRepo, repoInitialization, expandedProjectID, err := expandGitRepository(resourceData)
-
-	require.Nil(t, err)
-	require.NotNil(t, expandedGitRepo)
-	require.NotNil(t, expandedGitRepo.Id)
-	require.Equal(t, *expandedGitRepo.Id, repoID)
-	require.NotNil(t, expandedProjectID)
-	require.Equal(t, *expandedProjectID, projectID)
-	require.Nil(t, repoInitialization)
-}
-
-func TestGitRepo_FlattenExpandInitialization_RoundTrip(t *testing.T) {
-	projectID := uuid.New()
-	project := core.TeamProjectReference{Id: &projectID}
-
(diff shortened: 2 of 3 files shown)


EliiseS:
@xuzhang3 Should `initalized `

azuredevops/internal/service/git/resource_git_repository.go
```diff
@@ -394,9 +394,14 @@ func expandGitRepository(d *schema.ResourceData) (*git.GitRepository, *repoIniti
 
 	var initialization *repoInitializationMeta = nil
 	initData := d.Get("initialization").(*schema.Set).List()
-
-	// Note: If configured, this will be of length 1 based on the schema definition above.
-	if len(initData) == 1 {
+	if len(initData) == 0 {
```
EliiseS: @xuzhang3 @tmeckel Should initialization be a required field? I can't find a example where it's not set to be something,

xuzhang3: Not required. But if not set, terraform will crash #54

beandrad: This PR should fix that and default to `Uninitialized`. I think `Uninitialized` makes sense as a default, as the repo will be created with the default branch and no commits.

EliiseS: @beandrad Instead initializing it with default, let's make the `initialization` field required with 

```go
			"initialization": {
				Type:     schema.TypeSet,
				Required: true,
				MaxItems: 1,
```

tmeckel: I agree with @EliiseS , better force people to be explicit on what they want, instead for heading with a (potentially) unwanted default.

azuredevops/internal/service/git/resource_git_repository_test.go
```diff
@@ -121,7 +122,7 @@ func TestGitRepo_FlattenExpand_RoundTrip(t *testing.T) {
 	require.Equal(t, *expandedGitRepo.Id, repoID)
 	require.NotNil(t, expandedProjectID)
 	require.Equal(t, *expandedProjectID, projectID)
-	require.Nil(t, repoInitialization)
+	require.Equal(t, repoInitialization.initType, string(RepoInitTypeValues.Clean))
 }
 
 func TestGitRepo_FlattenExpandInitialization_RoundTrip(t *testing.T) {
```
EliiseS: Is this test testing anything different from the `TestGitRepo_FlattenExpand_RoundTrip` now? 

Maybe this? But could you move these to the above test?
```go
	require.Equal(t, repoInitialization.sourceType, "")
	require.Equal(t, repoInitialization.sourceURL, "")
```

Also, having the `configureCleanInitialization(resourceData)` come after the `flattenRepository`, seems incorrect. That wouldn't happen ever in a normal use case would it? @tmeckel

beandrad: oh yeah, they seem to be the same now; I guess I can just remove `TestGitRepo_FlattenExpand_RoundTrip` and rename `TestGitRepo_FlattenExpandInitialization_RoundTrip` as `TestGitRepo_FlattenExpand_RoundTrip`.

tmeckel: @EliiseS your're right the `flattenRepository` must come last! It's purpose is to serialize all settings into the Terraform state and if you change settings after this serialization the changes won't end up in the state.

azuredevops/internal/service/git/resource_git_repository_test.go
```diff
@@ -109,29 +109,6 @@ func TestGitRepo_FlattenExpand_RoundTrip(t *testing.T) {
...
-	projectID := uuid.New()
-	project := core.TeamProjectReference{Id: &projectID}
-
-	repoID := uuid.New()
-	repoName := "name"
-	gitRepo := git.GitRepository{Id: &repoID, Name: &repoName, Project: &project}
-
 	resourceData := schema.TestResourceDataRaw(t, ResourceGitRepository().Schema, nil)
 	resourceData.SetId(gitRepo.Id.String())
 	flattenGitRepository(resourceData, &gitRepo)
```
EliiseS: Can you switch this line with the one below it?

EliiseS:
Looks great! Thank you so much for your contribution!
Ship it 😄 
```
     _
    /|\
   /_|_\
 ____|____
 \_o_o_o_/
~~ |     ~~~~~
___t_________
````

PR #121, opened 2020-08-01
Add support to manage Teams in Projects

tmeckel:
## All Submissions:

This pull request contains the following new resources:

* `azuredevops_team` - for managing a Team
* `azuredevops_team_members` - for managing members of a team
* `azuredevops_team_administrators` - for managing administrators of a team

This pull request contains the following new data sources:

* `azuredevops_team` - to load information about a Team inside a Project in Azure DevOps
* `azuredevops_teams` - to load all Teams defined in a a Project in Azure DevOps or all (accessible) Teams of all (accessible) Projects

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information

#343


tmeckel:
@xuzhang3 finally I found some time to finish up the PR. Ready for review

tmeckel:
There is an issue with the  `tfproviderdocs` tool

image

Therefore the pipeline is failing at an early state. I can reproduce the error locally. So I guess the repo is kinda broken. @xuzhang3 shouldn't we not pin the tools which got installed with `GNUmakefile` to a specific version`?

copy @bflad and @katbyte for reference

xuzhang3:
Hi @tmeckel Fixed and merged, please update the latest code.

tmeckel:
@xuzhang3 will rebase the PR later today

xuzhang3:
@tmeckel   I have reverted the SDK upgrade. Current master code is based on v0.1.4,  the tools install issue has been fixed in PR #376  . Sorry for the inconvenient.

tmeckel:
@xuzhang3 No worries! Thanks for having the issues sorted and fixed! 👍🏼 👍🏼

tmeckel:
@xuzhang3 ping! 😀

xuzhang3:
Hi @tmeckel I will review this PR next week!

tmeckel:
@xuzhang3 while debugging the acceptance test I came across the **infamous** `'<' looking for beginning of value` error again! I guess we have a real issue here! Really sucks to debug, because it only shows up intermittently!

image

xuzhang3:
> @xuzhang3 while debugging the acceptance test I came across the **infamous** `'<' looking for beginning of value` error again! I guess we have a real issue here! Really sucks to debug, because it only shows up intermittently!
> 
> image

This error not always happen. A potential error is service returned the response as HTML(PAT expires or other errors), the [client.UnmarshalCollectionJson](https://github.com/microsoft/terraform-provider-azuredevops/blob/master/vendor/github.com/microsoft/azure-devops-go-api/azuredevops/client.go#L306) cannot marshal the response object and return the parse error.

An error example:
```go
func main() {
	var responseValue []build.DefinitionResourceReference
	err := json.Unmarshal([]byte("<html></html>"), &responseValue)
	fmt.Println(err)
}

Result:
invalid character '<' looking for beginning of value

```

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
```
xuzhang3: Service will only set description with empty string if not set, `Computed: true` can be removed. Also, we can set a default value for `description`. As far as I know, the description accept any string with length between 0~256

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+			},
+			"administrators": {
```
xuzhang3, line 47: Though API does not have the administrator number constraint ,  I would like to add a constraint that customer should add at least one administrator for the team same as portal did.
image

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+			return err
+		}
+	}
+
+	d.SetId(team.Id.String())
+	return resourceTeamRead(d, m)
+}
+
+func resourceTeamRead(d *schema.ResourceData, m interface{}) error {
+	log.Print("[DEBUG] resourceTeamRead: START")
```
xuzhang3: I think we can remove the operation progress log,  not helpful in bug trace without context information.

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+	return nil
+}
+
+func addTeamMembers(clients *client.AggregatedClient, team *core.WebApiTeam, query linq.Query) error {
+	idList, err := getIdentitiesFromSubjects(clients, query)
+	if err != nil {
+		return err
+	}
+	if idList == nil || len(*idList) != query.Count() {
+		return fmt.Errorf("Failed to load identity data for subjects")
```
xuzhang3, line 472: Return an error message with : `xxx cannot be found within this project or organization` might be better

azuredevops/internal/acceptancetests/resource_team_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	%s
+
+	resource "azuredevops_team" "team" {
+		project_id = azuredevops_project.project.id
+		name = "%s"
+		description = "%s"
+	}
+
+		`, projectResource, teamName2, teamDescription2)
```
xuzhang3: Extract the test data from this function to a separate test data function should be better.

azuredevops/internal/acceptancetests/resource_team_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
+)
+
+func TestAccTeam_CreateAndUpdate(t *testing.T) {
```
xuzhang3, line 14: Can we add more test case to cover the scenario like add admin/remove admin, add member/remove member. This test only cover the basic team create/update scenario.

azuredevops/internal/service/core/data_team.go
```diff
@@ -0,0 +1,35 @@
+package core
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/client"
+)
+
+func DataTeam() *schema.Resource {
+	baseSchema := ResourceTeam()
```
xuzhang3: I would suggest create the schema itself, not generate from resource schema. When the source schema changed, this won't effect the data source. Also some properties are not needed in data source.

azuredevops/internal/acceptancetests/data_team_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestAccTeam_DataSource(t *testing.T) {
+	projectName := testutils.GenerateResourceName()
+	projectResource := testutils.HclProjectResource(projectName)
+
+	config := fmt.Sprintf(`
+
+%s
+
+data "azuredevops_team" "team" {
```
xuzhang3, line 22: Extract the test data from this function to a separate test data function should be better.

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+				Set:        schema.HashString,
+			},
+			"members": {
+				Type: schema.TypeSet,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringIsNotWhiteSpace,
+				},
+				Optional:   true,
+				Computed:   true,
```
xuzhang3, line 65: Though my test, service won't add any member by default.  Can we remove the `Computed:true`?

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
...
+				Computed: true,
+			},
+			"administrators": {
+				Type: schema.TypeSet,
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringIsNotWhiteSpace,
+				},
+				Optional:   true,
+				Computed:   true,
```
xuzhang3, line 54: Though my test, service won't add any admin by default.  Can we remove the `Computed:true`?

azuredevops/internal/acceptancetests/resource_team_members_test.go
```diff
@@ -0,0 +1,251 @@
...
+data "azuredevops_group" "builtin_project_readers" {
+	project_id = azuredevops_project.project.id
+	name       = "Readers"
+}
+
+resource "azuredevops_team" "team" {
+	project_id = azuredevops_project.project.id
+	name = "%s"
+}
+
```
xuzhang3: Extract the test data from this function to a separate test data function should be better.

azuredevops/internal/acceptancetests/resource_team_administrators_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+
+	`, projectResource, teamName)
+
+	config2 := fmt.Sprintf(`
+
+%s
+
+data "azuredevops_group" "builtin_project_contributors" {
+	project_id = azuredevops_project.project.id
```
xuzhang3: Extract the test data from this function to a separate test data function should be better.

website/docs/r/team_administrators.html.markdown
```diff
@@ -0,0 +1,75 @@
...
+  project_id = azuredevops_project.project.id
+  name       = "${azuredevops_project.project.name} Team 2"
+}
+
+resource "azuredevops_team_administrators" "team_administrators" {
+  project_id     = azuredevops_team.team.project_id
+  team_id        = azuredevops_team.team.id
+  mode           = "overwrite"
+  administrators = [
+    azuredevops_group.builtin_project_contributors.descriptor
```
xuzhang3:
```suggestion
data.azuredevops_group.builtin_project_contributors.descriptor
```

azuredevops/internal/service/core/resource_team.go
```diff
@@ -0,0 +1,558 @@
+package core
```
xuzhang3, line 1: For functions like `setTeamAdministratorsPermissions`, `updateTeamAdministrators` etc. It would be better extract this functions into an Utils/helper file since these functions are used in different resources.

website/docs/d/data_team.html.markdown
```diff
@@ -0,0 +1,51 @@
...
+  name               = "Test Project"
+  work_item_template = "Agile"
+  version_control    = "Git"
+  visibility         = "private"
+  description        = "My first project"
+}
+
+data "azuredevops_team" "team" {
+	project_id = azuredevops_project.project.id
+	name = "${azuredevops_project.project.name} Team"
```
xuzhang3: Suggest change to `Team Project Team`, the simple the better.

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+}
+
+data "azuredevops_group" "builtin_project_readers" {
+  project_id = azuredevops_project.project.id
+  name       = "Readers"
+}
+
+resource "azuredevops_team" "team" {
+  project_id = azuredevops_project.project.id
+  name = "${azuredevops_project.project.name} Team 2"
```
xuzhang3: Suggest change to `My Test Team`,  plain text is better.

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+
+- `id` - The ID of the Team.
+
+## Relevant Links
+
+- [Azure DevOps Service REST API 5.1 - Teams - Create](https://docs.microsoft.com/en-us/rest/api/azure/devops/core/teams/create?view=azure-devops-rest-5.1)
+
+## Import
+
+Azure DevOps teams can be imported using the the complete resource id `<project_id>/<team_id>` e.g.
```
xuzhang3:
```suggestion
Azure DevOps teams can be imported using the complete resource id `<project_id>/<team_id>` e.g.
```

website/docs/r/team.html.markdown
```diff
@@ -0,0 +1,86 @@
...
+
+```sh
+$ terraform import azuredevops_team.id aed075ec-e178-47b6-96c2-314850128307/cd4be753-fda8-4107-94c9-59d9b808d35e
```
xuzhang3:
```suggestion
$ terraform import azuredevops_team.team 00000000-0000-0000-0000-000000000000/00000000-0000-0000-0000-000000000000
```

website/docs/d/data_teams.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+output "administrators" {
+  value = data.azuredevops_teams.test.teams.*.members
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `project_id` - (Optional) The Project ID. If no project ID all teams of all projects will ber returned.
```
xuzhang3:
```suggestion
- `project_id` - (Optional) The Project ID. If no project ID all teams of the organization will be returned.
```

xuzhang3 (changes requested):

azuredevops/internal/service/core/resource_team.go
tmeckel: Then you can't manage team administrators with the `azuredevops_team_administrators` resource.  Furthermore, I think that we should not always slavishly follow the functions in the web interface, since the group of users of the provider is different from the people who click around in the web page. We should make sure that Azure DevOps can be configured very flexible with the provider, even if that may mean breaking something, like having no administrator on a created team. What can be fixed easily anyway if discovered.

azuredevops/internal/service/core/resource_team.go
tmeckel: I agree on the `Computed` flag, but why should a default value be configured?

azuredevops/internal/service/core/resource_team.go
tmeckel: For such a detailed error message I must determine the difference between what's expected and what has been returned by the API. I'm not sure if it's really worth doing so, especially because when such a thing happens something is completely off inside Azure DevOps and thus you might need support from Microsoft to get it analyzed and fixed. Well at least I guess so.

azuredevops/internal/acceptancetests/resource_team_test.go
tmeckel: Yep should be feasible.

azuredevops/internal/service/core/data_team.go
tmeckel: ~~We have some other places were this is used as well on data sources that return information about `resources` implemented in the provider. So I think if the the resource schema changes, the data source schema should change as well at any time and using such an automatic way of deriving the data source schema from the "depended"  resource schema saves you from remembering to change the data source schema when you change the resource schema.~~

website/docs/d/data_team.html.markdown
tmeckel: See your point. 👍🏼

website/docs/r/team.html.markdown
tmeckel: 👍🏼

azuredevops/internal/service/core/resource_team.go
tmeckel: Aligned the code with `ResourceProject`

image

azuredevops/internal/acceptancetests/resource_team_test.go
tmeckel: Done => `testCreateTeamConfiguration`

azuredevops/internal/acceptancetests/resource_team_members_test.go
tmeckel: Done

azuredevops/internal/acceptancetests/resource_team_administrators_test.go
tmeckel: Done

azuredevops/internal/service/core/resource_team.go
tmeckel: The members are read anyway, so why not having the ability to process them at a later place inside a HCL?

azuredevops/internal/service/core/resource_team.go
tmeckel: See comment about `Compute` flag on `members`.

azuredevops/internal/service/core/data_team.go
tmeckel: Changed the code as proposed.

azuredevops/internal/acceptancetests/data_team_test.go
tmeckel: Done

azuredevops/internal/service/core/resource_team.go
tmeckel: Because the mentioned functions are only used within the scope of teams management and I don't expect that we have some sort of re-usage like with the helper functions in the `permission/utils` package, I'd like to keep the functions where they are located right now. But I see the point: when I look at the acceptance test files, that could be compiled and executed separately, this would definitely be an issue if (helper) functions are only located in one file that is not part of the compile unit. But as already said: I don't see that gonna happen here.

tmeckel:
@xuzhang3 I overhauled the code a bit 😀 ready for next review!

azuredevops/internal/acceptancetests/resource_team_test.go
tmeckel: Done 👍🏼

azuredevops/internal/service/core/resource_team.go
xuzhang3: I tested that we cannot set any string with a length greater than 256,  suggest add a length validation function for `description`: `ValidateFunc: validation.StringLenBetween(0, 256),`

azuredevops/internal/acceptancetests/resource_team_test.go
xuzhang3: Looks like the new code not submitted

azuredevops/internal/service/core/resource_team.go
xuzhang3: If member is configured in HCL,  the property here is not computed.  https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed

azuredevops/internal/acceptancetests/resource_team_test.go
tmeckel: @xuzhang3 I wrote the comment gefore I changed my mind again and moved the stuff to the `testutils/hcl.go` file as function `HclTeamConfiguration`. Sorry for the confusion.

azuredevops/internal/service/core/resource_team.go
tmeckel: Added 👍🏼 If the `description` is limited to a maximum number of characters, what about the `name`? I didn't find any description of the maximum length of the `name` of a Team.

azuredevops/internal/service/core/resource_team.go
tmeckel: If you read the description there's the following at the end `...and can be used as output for interpolation into other resources`. If you don't mark a property computed it cannot be used somewhere else. If you have a look in the AzureRM Terraform provider the tuple of `Optional: true` and `Computed: true` is all over the place. 

For example at the `Key Vault` resource for the `access_policy` property:

image

xuzhang3: Key vault will add at least one policy by default (user who created  it).  If the property value not set, service will auto set a default value for it, then this property can be `computed`.  Otherwise if we don't do anything and service won't do anything too, this property should not be an `computed` value.

tmeckel: No, the Key Vault resource does not create any access policies automatically:

image

image

That's because of the fact that there is a [`azurerm_key_vault_access_policy`](azurerm_key_vault_access_policy) to create access policies aside the key vault resource.

And if you create an access policy using the `azurerm_key_vault_access_policy`

image

You'll find this policy in the `key vault` resource because the `access_policy`attribute is `computed`

image

And thus you can use the `access_policy` from the `key_vault` resource in other calculations.

In the end I followed this strategy while I implemented the `azuredevops_teams` resource.

xuzhang3: 1. Currently the `accessPolicies`  is a required parameter:
   ```log
    {
        "error": {
            "code": "BadRequest",
            "message": "The parameter accessPolicies is not specified."
        }
    }
    ```
    I don‘t ’ know which API version used when key vault was introduced to AzureRM, from the AzureRM schema design that the `accessPolicies` used to be an optional but now it is required. AzureRM still works is because the request will auto set the `accessPolicies` to an empty array if `access_policy` not configured. [expandAccessPolicies](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/internal/services/keyvault/access_policy_schema.go#L175)
   
2. All configurations in `azurerm_key_vault` are in one API request,  key vault/access policies can be created or update with one request.

3. Back to ADO teams,   `admin` and `member` is not part of the team API,  you need call another APIs to manage admin or members, there is three different API here.   Combine these APIs as one resource is a good idea, treat them like one API.  But I still think there is not strong reason here that we should add `Computed: true` for `member`, `members` areismanaged by explicit configuration and API calls.

tmeckel: @xuzhang3 how can we reach an agreement here. I'd like to get this PR merged and it seems people are waiting for it actually.

tmeckel:
@xuzhang3 ping ❤️

xuzhang3:
@tmeckel  can you help resolve the conflicting files: `azuredevops/provider.go`

xuzhang3:
LGTM

xuzhang3 (approved):

PR #150, opened 2020-08-13
Location of golanci-lint is not always in gopath

sdebruyn:
golanci-lint can be installed by homebrew or as binary, so it's not always located inside gopath. It should be in the path however

## All Submissions:

/

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

/

## Other information
/


xuzhang3:
@sdebruyn  LGTM

PR #151, opened 2020-08-13
Do not set a default agent pool name

sdebruyn:
Only set the name of the agent pool if set by the user and don't set the default Ubuntu 16.04 pool. The pool is not required by the API so it shouldn't be required by the provider either.

## All Submissions:

The agent pool name was optional, but it tried to use Ubuntu 16.04 when the user didn't set it. This is not required by the API so the provider should not try to set the Ubuntu pool.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

/

## Other information
/

### DIFF
--- a/azuredevops/internal/service/build/resource_build_definition.go
+++ b/azuredevops/internal/service/build/resource_build_definition.go
@@ -325,7 +325,10 @@ func flattenBuildDefinition(d *schema.ResourceData, buildDefinition *build.Build
 	d.Set("name", *buildDefinition.Name)
 	d.Set("path", *buildDefinition.Path)
 	d.Set("repository", flattenRepository(buildDefinition))
-	d.Set("agent_pool_name", *buildDefinition.Queue.Pool.Name)
+
+	if buildDefinition.Queue != nil && buildDefinition.Queue.Pool != nil {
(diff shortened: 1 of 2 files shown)


xuzhang3:
Hi  @sdebruyn  [flattenBuildDefinition](https://github.com/terraform-providers/terraform-provider-azuredevops/blob/master/azuredevops/internal/service/build/resource_build_definition.go#L329) should be updated , otherwise when `pool` is not set,  provider will crash.

```
if buildDefinition.Queue != nil && buildDefinition.Queue.Pool != nil {
		d.Set("agent_pool_name", buildDefinition.Queue.Pool.Name)
	}
```

sdebruyn:
> Hi @sdebruyn [flattenBuildDefinition](https://github.com/terraform-providers/terraform-provider-azuredevops/blob/master/azuredevops/internal/service/build/resource_build_definition.go#L329) should be updated , otherwise when `pool` is not set, provider will crash.
> 
> ```
> if buildDefinition.Queue != nil && buildDefinition.Queue.Pool != nil {
> 		d.Set("agent_pool_name", buildDefinition.Queue.Pool.Name)
> 	}
> ```

@xuzhang3 done :)

xuzhang3:
Hi @sdebruyn  LGTM . Thanks for your contribution

PR #167, opened 2020-09-02
Adding tags to build pipeline

exocom:
## All Submissions:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] I have updated the documentation accordingly.
* [X] I have added tests to cover my changes.
* [X] All new and existing tests passed.
* [X] My code follows the code style of this project.
* [X] I ran lint checks locally prior to submission.
* [X] Have you checked to ensure there aren't other open PRs for the same update/change?

Adding the ability to set tags on build pipeline.

Issue Number:

- [ ] Yes
- [X] No

- [ ] Yes
- [x] No


exocom:
@xuzhang3 The tags schema is the same across multiple api endpoints (10+). I've placed it into **models** folder.

Is **models** a good place to put a shared schema?
I noticed other schemas are returned via a function. Is there any reason to wrap this in a function instead of exporting it directly? If so could you explain why?

xuzhang3:
@exocom  Some of the resources have the same base schema structure, so they are wrapped into util functions.  
It's OK to put the shared schema into **models** folder if tags is shared among different resources.

xuzhang3:
Hi @exocom  , Can you help add AccTest to cover the new  schema tags? I test with your branch, but tags cannot be added to the  build definition.   Looks like this is a service issues.

exocom:
@xuzhang3 Good catch, there is a bug on the API. I've submitted an issue to https://github.com/MicrosoftDocs/vsts-rest-api-specs/issues/429.

Also created an ACC test, however, waiting for feedback on the bug. Hopefully they can fix it otherwise will have to change the implementation.

Issue #59, opened 2020-07-06
Documentation for import of service endpoints is missing


nmiodice:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The documentation for how to import the following service endpoints are missing, but the code to do the import is working and tested:
* https://www.terraform.io/docs/providers/azuredevops/r/serviceendpoint_azurerm.html
* https://www.terraform.io/docs/providers/azuredevops/r/serviceendpoint_dockerregistry.html
* https://www.terraform.io/docs/providers/azuredevops/r/serviceendpoint_kubernetes.html

### New or Affected Resource(s)

* azuredevops_serviceendpoint_azurerm
* azuredevops_serviceendpoint_dockerregistry
* azuredevops_serviceendpoint_kubernetes

### Potential Terraform Configuration

N/A

### References

N/A

xuzhang3:
@nmiodice  I think that all of this resources contains sensitives data,  and the service always returns these variables with empty string. If these resource can be imported, there will be a diff during the plan stage, customers need to  handle the diffs manually. Support for importing resources containing sensitive configuration is inappropriate.

PR #191, opened 2020-09-24
Remove service connections import description

xuzhang3:
## All Submissions:

Issue Number: #59 
@nmiodice @EliiseS  I remove the service endpoints `import` description.  All service endpoints use PAT or password for authorization. If we support import this resource, sensitive data like PAT, password will be lost because service response always returned in empty string. If the the new user doesn't know the original PAT or password, they cannot reuse the the imported resource because the authorization will failed or he can use new PAT or password, but this make no difference with create a new service endpoint.  If users need the service endpoint name or ID, data source  can meet this requirement. This also happen to variable group if variable group contains secret values, we don't know whether the user  has the original data.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/website/docs/r/serviceendpoint_aws.html.markdown
+++ b/website/docs/r/serviceendpoint_aws.html.markdown
@@ -53,10 +53,3 @@ The following attributes are exported:
 ## Relevant Links
 * [aws-toolkit-azure-devops](https://github.com/aws/aws-toolkit-azure-devops)
 * [Azure DevOps Service REST API 5.1 - Agent Pools](https://docs.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpoints?view=azure-devops-rest-5.1)
-
-## Import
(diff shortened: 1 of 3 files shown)


xuzhang3:
@tmeckel ping you here, maybe you are interested in this issue 😄

tmeckel:
@xuzhang3 Interesting! I suspect that when you're talking in your comment at #59 about that values are missing, that values for variables that are considered secrets, like a password, app client secret or a certificate always be empty, when the details are read for a specific service endpoint using a `GET` operation on `/{organization}/{project}/_apis/serviceendpoint/endpoints/{endpointId}` and thus an import could not be successful because a required values, like the password, is missing for a HCL/State ServiceEndpoint definition. Right? 

If so, I've to say that, although I do share your concerns about importing resources with sensitive data, there should be special import functions to import existing ServiceEndpoint definitions, so that the provider is really valuable for people and doesn't have a lot of obstacles. The import functions should support specifying the required login/secret data (values aren't returned from the GET operation) to fully configure the existing ServiceEndpoint in the provider.

Issue #192, opened 2020-09-24
terraform corrupted state: with Error: rpc error: code = Unavailable desc = transport is closing 


poligraph:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

terraform -v
Terraform v0.12.29
+ provider.azuredevops v0.0.1
+ provider.azurerm v2.28.0
+ provider.google v3.11.0

### Affected Resource(s)

* `azuredevops_branch_policy_build_validation.build-validation`

### Terraform Configuration Files

```hcl
resource "azuredevops_git_repository" "repo" {
  project_id = var.project-id
  name       = var.repo-name
  initialization {
    init_type = "Clean"
  }

  lifecycle {
    prevent_destroy = true
  }

}

resource "azuredevops_build_definition" "build" {
  count      = var.create-master-pipeline ? 1 : 0
  project_id = var.project-id
  name       = var.pipeline-name != "false" ? var.pipeline-name : var.repo-name
  path       = "\\${var.repo-folder}"

  ci_trigger {
    use_yaml = true
  }

  repository {
    repo_type = "TfsGit"
    repo_id   = azuredevops_git_repository.repo.id
    yml_path  = "pipeline.yml"
  }
}

resource "azuredevops_branch_policy_build_validation" "build-validation" {
  count      = var.enable-pr-pipeline ? 1 : 0
  project_id = var.project-id

  enabled  = var.enable-pr-pipeline
  blocking = true

  settings {
    display_name        = "check-pr"
    build_definition_id = azuredevops_build_definition.build[count.index].id
    valid_duration      = 720

    scope {
      repository_id  = azuredevops_git_repository.repo.id
      repository_ref = azuredevops_git_repository.repo.default_branch
      match_type     = "Exact"
    }
  }
}
```

### Debug Output

https://gist.github.com/poligraph/7fdec0ec3bd43118d86b5d1d83ddc3e0

### Expected Behavior

terraform plan should work

### Actual Behavior

```
Error: rpc error: code = Unavailable desc = transport is closing
[...]

Issue #123, opened 2020-08-03
azuredevops_variable_group crashes on import


A30004028:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

* Terraform v0.12.26
* provider.azuredevops v0.0.1

### Affected Resource(s)

* `azuredevops_variable_group`

### Terraform Configuration Files

```hcl
data "azuredevops_project" "project" {
  project_name = "AAA"
}

resource "azuredevops_variable_group" "main" {
  project_id   = data.azuredevops_project.project.id
  name         = "terraform-enterprise-test"
  description = "test"
  allow_access = true
  variable {
      name         = "terraform-team-token"
      secret_value        = "test"
      is_secret    = true
  }
}
```

### Debug Output
https://gist.github.com/A30004028/211299a02a409303cdad651145173019

### Panic Output

https://gist.github.com/A30004028/211299a02a409303cdad651145173019

### Expected Behavior

Resource is imported succesfully as per https://www.terraform.io/docs/providers/azuredevops/r/variable_group.html#import

### Actual Behavior

Resource import crashes

### Steps to Reproduce

1. Ensure an Azure Devops variable group exists in Azure DevOps with at least one variable, I used a sensitive value but it does not appear to affect the result.
1. Use terraform to import the variable group with the following, replacing the variable group ID with your own.
    `terraform import azuredevops_variable_group.main "AAA"/846`
   
### References

* #0000

xuzhang3:
Hi @A30004028 Thanks for your feedback.   This is because you have secret values in your configuration,  service won't return secret values. We fix this to support secret values improt ,but you will lost the secret values.

A30004028:
Thats fine, the secrets are maintained elsewhere in code. Am I right in thinking the values will get reapplied on the next plan/apply?

xuzhang3:
yes, you are correct.

DamonStamper:
[...]

PR #196, opened 2020-09-28
Import for secret variables

xuzhang3:
## All Submissions:

Issue Number: #192 #123 
For secret variable's value, service will always returned in empty string. If we create resource with secret variables, we need to read the secret values from the `.tfstate`. But if we try import resources with secret values, we cannot read any data from `.tfstate`, create a default variables can fix this issue, users can handle the secret variable after import finished.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/build/resource_build_definition.go
+++ b/azuredevops/internal/service/build/resource_build_definition.go
@@ -362,14 +362,19 @@ func flattenBuildVariables(d *schema.ResourceData, buildDefinition *build.BuildD
 	index := 0
 	for varName, varVal := range *buildDefinition.Variables {
 		var variable map[string]interface{}
-		if converter.ToBool(varVal.IsSecret, false) {
-			variable = tfhelper.FindMapInSetWithGivenKeyValue(d, bdVariable, bdVariableName, varName)
-		} else {
-			variable = map[string]interface{}{
-				bdVariableName:          varName,
-				bdVariableValue:         converter.ToString(varVal.Value, ""),
-				bdVariableIsSecret:      false,
-				bdVariableAllowOverride: converter.ToBool(varVal.AllowOverride, false),
+
+		isSecret := converter.ToBool(varVal.IsSecret, false)
+		variable = map[string]interface{}{
+			bdVariableName:          varName,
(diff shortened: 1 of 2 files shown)


azuredevops/internal/service/build/resource_build_definition.go
```diff
@@ -362,14 +362,19 @@ func flattenBuildVariables(d *schema.ResourceData, buildDefinition *build.BuildD
...
+		variable = map[string]interface{}{
+			bdVariableName:          varName,
+			bdVariableValue:         converter.ToString(varVal.Value, ""),
+			bdVariableIsSecret:      isSecret,
+			bdVariableAllowOverride: converter.ToBool(varVal.AllowOverride, false),
+		}
+
+		//read secret variable from state if exist
+		if isSecret {
+			if stateVal := tfhelper.FindMapInSetWithGivenKeyValue(d, bdVariable, bdVariableName, varName); stateVal != nil {
```
EliiseS, line 376: Is it possible to merge this if and the above if?

azuredevops/internal/service/taskagent/resource_variable_group.go
```diff
@@ -449,14 +449,21 @@ func flattenVariable(d *schema.ResourceData, variableAsJSON []byte, varName stri
...
 		vgValue:    converter.ToString(variable.Value, ""),
-		vgIsSecret: false,
-	}, nil
+		vgIsSecret: isSecret,
+	}
+
+	//read secret variables from state if exist
+	if isSecret {
+		stateVal := tfhelper.FindMapInSetWithGivenKeyValue(d, vgVariable, vgName, varName)
+		if stateVal != nil {
```
EliiseS: Move the variable declaration to this line, to be consistent with the if statement in  `resouce_build_def.go`

azuredevops/internal/service/build/resource_build_definition.go
xuzhang3: No , I don't think the two if statement can be merged.
The logic here is :
If the variable is not secret, there's no need to read it from state.
If the variable is secret,  try read the value from state . If  can be read, the state value will be returned otherwise an instance build with service response will be returned.

PR #200, opened 2020-10-06
Search for Project Collection Groups with Group datasource

tmeckel:
## All Submissions:

The PR contains a change to the Group datasource , so that it's possible to search for project collection groups and not only for groups located inside a project.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/graph/data_group.go
+++ b/azuredevops/internal/service/graph/data_group.go
@@ -19,15 +19,14 @@ func DataGroup() *schema.Resource {
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:         schema.TypeString,
-				ForceNew:     true,
 				Required:     true,
-				ValidateFunc: validation.NoZeroValues,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 			},
 			"project_id": {
 				Type:         schema.TypeString,
-				ForceNew:     true,
-				Required:     true,
-				ValidateFunc: validation.NoZeroValues,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 			},
 			"descriptor": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


website/docs/d/data_group.html.markdown
```diff
@@ -28,14 +28,27 @@ output "group_id" {
...
+
 ```
 
 ## Argument Reference
 
 The following arguments are supported:
 
-- `project_id` - (Required) The Project Id.
 - `name` - (Required) The Group Name.
+- `project_id` - (Optional) The Project Id. If no project Id is specified the project collection groups will be searched.
```
xuzhang3: I would suggest change `Id` to `ID`

azuredevops/internal/service/graph/data_group.go
```diff
@@ -19,15 +19,14 @@ func DataGroup() *schema.Resource {
...
-				ValidateFunc: validation.NoZeroValues,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
 			},
 			"project_id": {
 				Type:         schema.TypeString,
-				ForceNew:     true,
-				Required:     true,
-				ValidateFunc: validation.NoZeroValues,
+				Optional:     true,
+				Computed:     true,
```
xuzhang3: `project_id` is required when we need to read project groups.  Both of  project groups and collection groups won't set `project_id` when service response returned. I think  we can remove `Computed:     true,` as this configuration is configured by the users.

xuzhang3 (changes requested):
Few changes requested but otherwise looks good

website/docs/d/data_group.html.markdown
tmeckel: Done

azuredevops/internal/service/graph/data_group.go
tmeckel: Yes I agree, you're right. I removed the `computed` flag for `project_id`

xuzhang3:
LGTM

Issue #404, opened 2021-06-15
azuredevops_git_permissions cannot set Git folder permissions


timolbg:
The resource azuredevops_git_permissions doesn't allow setting "create branch" permissions on a folder. This is commonly needed to restrict Contributors to create branches only in feature/*, bugfix/*, etc.

I get the error message:
```
Error: No branch found with name [refs/heads/feature] in repository with id [c7451565-b51d-4189-972f-1a02db019993]

  on modules/fsm-repo/main.tf line 70, in resource "azuredevops_git_permissions" "creat_branch_feature_allow":
  70: resource "azuredevops_git_permissions" "creat_branch_feature_allow" {
```

However, this is possible using the "az devops permission" commandline tool. So I don't think this is a limitation of the RESTful API, but caused by the terraform resource implementation:
```
# Deny "create branch for Contributor for root branch:
az devops security permission update --id $namespaceId --org $orgId --subject $descriptor.descriptor --token "repoV2/$projectid/$repoId" --deny-bit 16 --allow-bit 16494
# Allow "create branch" on feature/*
$hexFeatureBranch = hexify -string  "feature"
$featureToken = "refs/heads/$hexFeatureBranch"
$featureTokenBuild = "repoV2/$projectid/$repoid/$featureToken"
az devops security permission update --id $namespaceId --org  $orgId --subject $descriptor.descriptor --token $featureTokenBuild --deny-bit 0 --allow-bit 16
```
The limitation is seemingly in the section of the code where you are validating the branch name:
https://github.com/microsoft/terraform-provider-azuredevops/blob/18a3d25288dff8437a6089ebc14932dc0ef0e3a8/azuredevops/internal/service/permissions/resource_git_permissions.go#L124

Folders are not listed by "az repos ref list".
If validating the branch_name became optional, it would not fail anymore as the aclToken would be generated properly and the rest of the code would execute normally. I understand that this might not be desirable, hence introducing a new parameter would be better.

Thanks!

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azuredevops_XXXXX

### Potential Terraform Configuration

[...]

PR #411, opened 2021-06-27
Do not check branch existence for git permission

Marc3001:
## All Submissions:

According to https://github.com/microsoft/terraform-provider-azuredevops/issues/404 azure DevOps API allows creating branch permissions even if branch does not exist.
Resource actually does not allow it.

In issue we first discussed about adding attribute `token` which will modify behaviour (bypass branch check) and let current behaviour if `branch_name` attribute is used.
It could lead to conflicts as we could have some resources with `token` value same as another with `branch_name`. So this change is finally just always bypassing branch existence and do not add any new attribute. It will not break existing conf and behaviour will be closer to azure DevOps API behaviour.

Also I checked the documentation and I didn't find anywhere that branch should exist so I didn't change anything in it.

Issue Number:
https://github.com/microsoft/terraform-provider-azuredevops/issues/404

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Nope

## Other information
Nope


xuzhang3:
@Marc3001 Thanks for your PR. Looks good to me . Not sure if service will enhance the branch name checking in the future but it's ok for now.
Acceptance test result:
```log
=== RUN   TestAccGitPermissions_SetPermissions
=== PAUSE TestAccGitPermissions_SetPermissions
=== CONT  TestAccGitPermissions_SetPermissions
--- PASS: TestAccGitPermissions_SetPermissions (91.07s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        92.168s

```

PR #412, opened 2021-06-29
change the apiserver_url description

ahmedig:
The description is very misleading. It says the service connection description, which is not the case. Its the Kube API's hostname.


xuzhang3:
LGTM

Issue #491, opened 2021-11-30
Build definition import fails on invalid memory address


roryprimrose:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform: 1.0.11
Azure DevOps provider: 0.1.7

### Affected Resource(s)

* `azuredevops_build_definition`

### Terraform Configuration Files

```hcl
resource "azuredevops_build_definition" "pipeline" {
  project_id      = var.repository.project_id
  name            = var.config.name
  path            = var.config.folderPath
  agent_pool_name = var.config.pool

  ci_trigger {
    use_yaml = true
  }

  repository {
    repo_type           = "TfsGit"
    repo_id             = var.repository.id
    branch_name         = var.repository.default_branch
    yml_path            = var.config.yamlPath
    report_build_status = var.config.reportBuildStatus
  }
}

```

### Debug Output

```
module.azdo_pipeline["7d625088-2797-4329-a50c-e6c452456678"].azuredevops_build_definition.pipeline: Import prepared!
  Prepared azuredevops_build_definition for import
module.azdo_pipeline["7d625088-2797-4329-a50c-e6c452456678"].azuredevops_build_definition.pipeline: Refreshing state... [id=41]
╷
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵


Stack trace from the terraform-provider-azuredevops_v0.1.7 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xeaaf37]

goroutine 39 [running]:
github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/service/build.flattenRepository(0xc000192160, 0x11c9cf9, 0x4)
	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/service/build/resource_build_definition.go:510 +0x3d7
[...]

Issue #204, opened 2020-10-15
Agent Pool - Azure virtual machine scale set


aleqsss:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
The 'Azure virtual machine scale set' pool type of agent pools, inside of Azure DevOps, has now been released and is publicly available, since the (what I can see) version 6.0 of the Azure DevOps REST API. This makes me think that it would be suitable to be able to configure such, from this provider. 

As you can see in the documentation, it now implementes the [TaskAgentPoolOptions](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/pools/add?view=azure-devops-rest-6.0#taskagentpooloptions) definition which allows you to specify, amongst others, the **elasticPool** property.

### New or Affected Resource(s)

Affected Resources:
* azuredevops_agent_pool

Affected Data Sources:
* azuredevops_agent_pool
* azuredevops_agent_pools

### References
From what I can see, this provider is using the [azure-devops-go-api](https://github.com/microsoft/azure-devops-go-api) which, at the time of writing, doesn't implement the **elasticPool** / [TaskAgentPoolOptions](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/pools/add?view=azure-devops-rest-6.0#taskagentpooloptions). Because of this I've created the following issue in that repository: 

* [Add new taskagent resource functionality #90](https://github.com/microsoft/azure-devops-go-api/issues/90)

aleqsss:
I'm interested in working/contributing on this issue, but I'm impeded because as of my understanding, the [azure-devops-go-api](https://github.com/microsoft/azure-devops-go-api) is using an internal tool/code generator to generate the SDK.

simongottschlag:
@xuzhang3 would it be possible for you to ask internally at Microsoft what the plans are for the updated SDK?

@aleqsss has been waiting for a reply for quite some time and would be great to understand what the plans are.

I created the following more than a week ago and still nothing: https://github.com/microsoft/azure-devops-go-api/issues/93

Thank you 😊

xuzhang3:
[...]

Issue #143, opened 2020-08-10
[Feature request] Add support for "pipeline" > "Environments"


DamonStamper:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add support for a azuredevops_environment resource.

Ideally support defining an environment, ignore deploys made to it, and defining "approvals and checks".

### New or Affected Resource(s)

* azuredevops_environment

xuzhang3:
Hi @DamonStamper  , Environments API is supported at v6.0,  Current sdk API version is 5.1. This resource will pending until **auzde-deviops-sdk** support it. [SDK issue](https://github.com/microsoft/azure-devops-go-api/issues/86)

Marcus-James-Adams:
Is there a time scale for v6.0 of the SDK?  Our pipelines rely heavily on this.

xuzhang3:
Azure-DevOps-GO-SDK is managed by another team,  they have working on it, but don't have a release date.

DevOpsBoondoggles:
This would definitely be great since it's the only way to do manual approval stages via yaml. Appreciate the dependency is not from you though.

picardsrcd:
> This would definitely be great since it's the only way to do manual approval stages via yaml. Appreciate the dependency is not from you though.

An interim approach is possible with null-resource 
https://joefecht.com/posts/creating-an-azdo-environment-via-api/


```hcl
// Create Environment using API
// https://github.com/microsoft/terraform-provider-azuredevops/issues/143
resource "null_resource" "environment" {
  provisioner "local-exec" {
    command = "./create_environment.sh"

    environment = {
      azdo_pat_token_owner_email = var.azdo_pat_token_owner_email
      azdo_pat_token             = var.azdo_pat_token
      project_url                = module.azdo.azdo_project_url

      name = "core_infrastructure_env"
    }
  }
}
```

```bash
# This script to call the API is required because of
# https://github.com/microsoft/terraform-provider-azuredevops/issues/143
# The Go SDK is not updated yet to interact with the API V6.0

# This is based on content from https://joefecht.com/posts/creating-an-azdo-environment-via-api/
[...]

PR #494, opened 2021-12-03
Upgrade to v6 API

tmeckel:
## All Submissions:

* [ ] ~~I have updated the documentation accordingly.~~
* [ ] ~~I have added tests to cover my changes.~~

Issue Number: #204 #143 #491

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


tmeckel:
@nechvatalp I successfully migrated the provider to the v6 API. 

One **major issue** though: it is impossible to set a description for a Service Endpoint using [ServiceEndpointClient.CreateServiceEndpoint](https://docs.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpoints/create?view=azure-devops-rest-6.0)!

I verified this behavior using a Python script which utilizes the v6 Python API:

```Python
credentials = BasicAuthentication('', personal_access_token)
connection = Connection(base_url=organization_url, creds=credentials)

cc: Type[CoreClient]
cc = connection.clients_v6_0.get_core_client()
prj: Type[TeamProject]
prj = cc.get_project("Test Project")

spc: Type[ServiceEndpointClient]
spc = connection.clients_v6_0.get_service_endpoint_client()

ep_name = "Generic (TEST)"
ep = ServiceEndpoint(
    name=ep_name,
    description="Test description",
    type="generic",
    url="https://void.example.com",
    authorization=EndpointAuthorization(
        parameters={
            "username": "test",
            "password": "test"
        },
        scheme="UsernamePassword"),
    service_endpoint_project_references=[
        ServiceEndpointProjectReference(
            project_reference=ProjectReference(
                id=prj.id
            ),
            name=ep_name
        )
    ])

response = spc.create_service_endpoint(ep)
```

The description got passed via the `POST` Request to the API, but never got set. This can be easily verified by selecting the service endpoint of the corresponding project in the web interface. 

image

This behavior will always create a DIFF in Terraform because the `description` attribute is always empty.

We have three options here:

1. Staying with v5.1 API for the Service Endpoint. Definitely not my favorite because personally I really don't like a mixture of APIs inside such a project like the provider.
2. Ignoring changes on the description property entirely. 

    image

    Likewise with the v5.1 API not a great option, because this would be a step back for the functionality of the provider.
3. Removing the description property for a Service Endpoint in the provider. I think the worst option because this would be a breaking change for everyone who use the provider for managing Service Endpoints.
4. Get the v6 API fixed 😁 My favorite solution!!!!!

@xuzhang3 what's you opinion on this?

favoretti:
Hey @tmeckel. Am I reading this wrong? https://github.com/microsoft/azure-devops-go-api/blob/0974d0145be828d8e35ef1b06a542e00938713ce/azuredevops/v6/serviceendpoint/models.go#L432

The ServiceEndpoint object does have a Description property...

favoretti:
@nechvatalp On thing I'm pondering though.. Why does this SDK use `6.0-preview.4` instead of 6.1 which seems to be used everywhere else?

roryprimrose:
@tmeckel The api model is a bit of a misdirect in this case. I find the easiest way to verify behaviour is to watch the browser traffic when the UI makes the change. Unfortunately the UI often does not use the formal API but in this case it does. The answer here is that the description property on the root object is not what you want to use.

image

image

The description in the UI is pushed into the object in the `serviceEndpointProjectReferences` array.

image

xuzhang3:
@tmeckel   I prefer  options 4. This is a feature in UI, API should not breaking this experience.
@roryprimrose   You are right, UI use API v`6.0-preview.4`  and the description has been wrapped in the `serviceEndpointProjectReferences`. This should be a breaking between v5 and v6.

tmeckel:
> @tmeckel The api model is a bit of a misdirect in this case. I find the easiest way to verify behaviour is to watch the browser traffic when the UI makes the change. Unfortunately the UI often does not use the formal API but in this case it does. The answer here is that the description property on the root object is not what you want to use.
> ...
> The description in the UI is pushed into the object in the `serviceEndpointProjectReferences` array.

@roryprimrose great catch! 👍🏼 I'll update the code accordingly.

tmeckel:
@xuzhang3 I consider this PR ready to be merged. I ran a couple of acceptance tests, especially for the Service Endpoints and the changed implementation for the User Entitlement resource

image

image

image

xuzhang3:
LGTM, shipping it

PR #536, opened 2022-02-15
Remove "ConflictsWith" to fix broken reset all branch policy

bermo:
## All Submissions:

The provider currently does not build a minimum reviewers branch policy properly if the "Reset all code reviewer votes" option is selected - if new changes are pushed, existing "Approved" votes are not removed.

Having a closer look at the rest api results for a valid policy using "Reset all code reviewer votes" shows the following settings:

``` json
"settings": {
        "minimumApproverCount": 1,
        "creatorVoteCounts": false,
        "allowDownvotes": false,
        "resetOnSourcePush": true,
        "requireVoteOnLastIteration": false,
        "resetRejectionsOnSourcePush": true,
        "blockLastPusherVote": true,
        "scope": [
          {
            "refName": "refs/heads/main",
            "matchKind": "Exact",
            "repositoryId": "306e0555-774a-4952-b729-e0607c4a49c0"
          }
        ]
      }
```
Notice that the both the `resetOnSourcePush` and `resetRejectionsOnSourcePush` are both true at the same time in this case ...

This seems to indicate that the `ConflictsWith: resetOnSourcePush` and `ConflictsWith: resetRejectionsOnSourcePush` tags are actually incorrect and causing the problem.

I am proposing to remove the `ConflictsWith` tags, but there might be a better way as I am not very familiar with go. ...

Issue Number:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/policy/branch/resource_branchpolicy_min_reviewers.go
+++ b/azuredevops/internal/service/policy/branch/resource_branchpolicy_min_reviewers.go
@@ -16,9 +16,9 @@ type minReviewerPolicySettings struct {
 	ApprovalCount                     int  `json:"minimumApproverCount" tf:"reviewer_count"`
 	SubmitterCanVote                  bool `json:"creatorVoteCounts" tf:"submitter_can_vote"`
 	AllowCompletionWithRejectsOrWaits bool `json:"allowDownvotes" tf:"allow_completion_with_rejects_or_waits"`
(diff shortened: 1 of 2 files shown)


xuzhang3:
[bermo](https://github.com/bermo)   Thanks for your feedback, this issue looks like the options: `resetOnSourcePush, requireVoteOnLastIteration, resetRejectionsOnSourcePush` are combination from service side not solo configuration as we see from the web page.  We should keep the behavior as the web page does, not leave it to users.

PR #626, opened 2022-08-03
fix: flatten project falls back to organization default template ID

### LINKED ISSUE #199: Error parsing Work Item Template ID, got : invalid UUID length: 0
### Community Note

```
PS /Users/timja/projects/hmcts/ops-bootstrap/Terraform/Environments/Sandbox> terraform -v
Terraform v0.13.2
+ provider registry.terraform.io/hashicorp/azuread v0.11.0
+ provider registry.terraform.io/hashicorp/azurerm v2.15.0
+ provider registry.terraform.io/hashicorp/random v2.3.0
+ provider registry.terraform.io/terraform-providers/azuredevops v0.0.1

Your version of Terraform is out of date! The latest version
is 0.13.4. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Resource(s)

* `data.azuredevops_project`

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

```
data "azuredevops_project" "project" {
  project_name = "DevOps"
}
```

### Debug Output

can't see anything useful there

### Expected Behavior

data source works

### Actual Behavior

```
Error: Error flattening project: Error parsing Work Item Template ID, got : invalid UUID length: 0

  on ../../Modules/Infrastructure/Management/main.tf line 323, in data "azuredevops_project" "project":
 323: data "azuredevops_project" "project" {
```

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

This project wasn't created by AzureDevOps, looks like it's expecting something to be setup that isn't.

### References

* #0000

wwmoraes:
The Azure DevOps API does not guarantee a process template ID will be returned when getting a project detail. This seems to be caused for projects which use the organisation default template. As such, the lookup logic should allow to fallback to that value, if none is present at the project level.

The changes introduced allow an empty work item template parameter, which will force a lookup on the organisation level. The default value is kept as `Agile` for compatibility and design choices.

Fixes #199

## All Submissions:

Issue Number:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/core/resource_project.go
+++ b/azuredevops/internal/service/core/resource_project.go
@@ -78,7 +78,6 @@ func ResourceProject() *schema.Resource {
 				Type:             schema.TypeString,
 				ForceNew:         true,
 				Optional:         true,
-				ValidateFunc:     validation.StringIsNotWhiteSpace,
 				DiffSuppressFunc: suppress.CaseDifference,
 				Default:          "Agile",
 			},
@@ -416,10 +415,20 @@ func deleteProject(clients *client.AggregatedClient, id string, timeoutSeconds t
 
 // Convert internal Terraform data structure to an AzDO data structure
 func expandProject(clients *client.AggregatedClient, d *schema.ResourceData, forCreate bool) (*core.TeamProject, error) {
-	workItemTemplate := d.Get("work_item_template").(string)
-	processTemplateID, err := lookupProcessTemplateID(clients, workItemTemplate)
-	if err != nil {
-		return nil, err
+	var processTemplateID string
+	var err error
+	workItemTemplate := strings.TrimSpace(d.Get("work_item_template").(string))
+	if len(workItemTemplate) > 0 {
+		processTemplateID, err = lookupProcessTemplateID(clients, workItemTemplate)
+		if err != nil {
+			return nil, err
+		}
+	} else { // use the organization default template if an empty string is set
+		processTemplateUUID, err := getDefaultProcessTemplateID(clients)
+		if err != nil {
+			return nil, err
+		}
+		processTemplateID = processTemplateUUID.String()
 	}
 
 	// an "error" is OK here as it is expected in the case that the ID is not set in the resource data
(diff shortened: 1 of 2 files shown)


xuzhang3:
@wwmoraes  this issue is still not fixed,  fallback to ORG default template may set the wrong template result in a state `diff`.

wwmoraes:
> @wwmoraes this issue is still not fixed, fallback to ORG default template may set the wrong template result in a state `diff`.

Thank you @xuzhang3 for the review! I've updated the code to remove the hardcoded `Agile` template and instead use the organisation default. This means a breaking change for a subset of existing customers that both:

* don't set `work_item_template` explicitly on their resources
* have a work item template set on their organisation different than `Agile`

This also has the side-effect that future changes on the organisation default template will cause all resources without `work_item_template` explicitly set to trigger a recreation. For such cases, the best practice is to set it.

xuzhang3:
@wwmoraes  I do not think change the default value to empty string and set it to the default organization process template is the best choice 
1. The default value will be changed via the organization default process template.
2. The default is not clear to users, it will set to the default process template of the organization,  not the empty string defined in the schema.
3. Breaking changes. The template cannot be found, this could be a permission issue not the APIs, though this fix can be a workaround for this issue.

wwmoraes:
@xuzhang3 thank you again for reviewing it 😄 

> The default value will be changed via the organization default process template.

This is exactly how the upstream service works: the organization itself has a default. Having a custom hardcoded value here means this provider has a distinct behaviour than the upstream service, which increases cognitive load. Mitigating an eventual recreation diff seems like a poor reason to do that. Besides, infrastructure-as-code is meant to be declarative, so a customer is prone to uncontrolled changes for things they do not explicitly configure, which seems fair.

> The default is not clear to users, it will set to the default process template of the organization, not the empty string defined in the schema.

I've updated the docs mentioning that the default is the one set on the organization. The empty string on the schema is due to the "offline" approach used for its creation. We could change the schema during initialization using the configure context function, but that wouldn't solve the declared empty string you're referring to.

If you mean the upstream service documentation is unclear about this then that's a problem of the upstream service provider.

> Breaking changes. The template cannot be found, this could be a permission issue not the APIs, though this fix can be a workaround for this issue.

If you mean the original issue, #199, then its not a permission issue: one of the users that tested it is the organization owner.

xuzhang3:
@wwmoraes  thanks for your feedback, I would like to keep the default type to `Agile` not set it to empty string and fall back to organization default template. Keep it clear and visible to users when run `terraform plan`. Agree with fall back to organization  default template when template not found.

wwmoraes:
@xuzhang3 would you be so kind as to review it again? I've reinstated the `Agile` default value, and configured an empty string to trigger the organisation fallback. 😄

xuzhang3 (approved):

xuzhang3:
@wwmoraes LGTM 🚢

Issue #692, opened 2023-01-18
azuredevops_serviceendpoint_npm not triggering a change in plan when a new access_token is provided (if the first 71 characters are the same)


liyaoz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version
Latest as of 18/01/2023: Terraform v1.3.7, AzDO provider v0.3

### Affected Resource(s)

* `azuredevops_serviceendpoint_npm`
* potentially other `azuredevops_serviceendpoint` as well.

### Terraform Configuration Files

```hcl
resource "azuredevops_serviceendpoint_npm" "artifactory_npm" { 
  project_id            = "xxxxxx"
  service_endpoint_name = "Artifactory-NPM"
  description           = "NPM service endpoint to artifactory"
  url                   = "https://does.not.matter/"
  access_token          = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAXXXXXXXXXXXXXXXXXXXXXX" 
  # -> replace 'X' with anything, remove or add more does not trigger a change in plan
}
```

### Expected Behavior

Terraform plan changes the service connection when a new 'access_token' is supplied

### Actual Behavior

```
> terraform plan
azuredevops_serviceendpoint_npm.artifactory_npm: Refreshing state...

No changes. Your infrastructure matches the configuration.
```

The Jfrog Artifactory produces an 809 characters access token, and the first 71 characters, unfortunately, does not change for a same user. This causes the service connection to stop working.

### Steps to Reproduce

1. `terraform apply`

### References

* #0000

davidcorrigan714:
Just hit this too interestingly enough. Trying to figure out where that 71 character limit is coming from. I'm working on adding the JFrog v2 connections and really thought I had messed something up when testing our rotation logic.

davidcorrigan714:
Here's the culprit: https://github.com/golang/go/issues/36546

davidcorrigan714:
[...]

PR #708, opened 2023-02-02
[Bug Fix] - `azuredevops_serviceendpoint_npm` Remove token hash

xuzhang3:
## All Submissions:

Remove `access_token` hash

Issue Number: #692 

```log
=== PAUSE TestAccServiceEndpointNpm_complete
=== RUN   TestAccServiceEndpointNpm_update
=== PAUSE TestAccServiceEndpointNpm_update
=== RUN   TestAccServiceEndpointNpm_RequiresImportErrorStep
=== PAUSE TestAccServiceEndpointNpm_RequiresImportErrorStep
=== CONT  TestAccServiceEndpointNpm_basic
=== CONT  TestAccServiceEndpointNpm_update
=== CONT  TestAccServiceEndpointNpm_RequiresImportErrorStep
=== CONT  TestAccServiceEndpointNpm_complete
--- PASS: TestAccServiceEndpointNpm_complete (48.02s)
--- PASS: TestAccServiceEndpointNpm_RequiresImportErrorStep (51.90s)
--- PASS: TestAccServiceEndpointNpm_basic (58.12s)
--- PASS: TestAccServiceEndpointNpm_update (86.36s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        86.894s

```

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/acceptancetests/resource_serviceendpoint_npm_test.go
+++ b/azuredevops/internal/acceptancetests/resource_serviceendpoint_npm_test.go
@@ -53,7 +53,6 @@ func TestAccServiceEndpointNpm_complete(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testutils.CheckServiceEndpointExistsWithName(tfSvcEpNode, serviceEndpointName),
 					resource.TestCheckResourceAttrSet(tfSvcEpNode, "project_id"),
-					resource.TestCheckResourceAttrSet(tfSvcEpNode, "access_token_hash"),
 					resource.TestCheckResourceAttr(tfSvcEpNode, "url", "https://url.com/"),
 					resource.TestCheckResourceAttr(tfSvcEpNode, "service_endpoint_name", serviceEndpointName),
 					resource.TestCheckResourceAttr(tfSvcEpNode, "description", description),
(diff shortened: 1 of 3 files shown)


davidcorrigan714:
@xuzhang3 Why are you removing the token hash? If you've decided that masking secret data in state is no longer an objective of this provider then it should be all or nothing. There's also other areas of the code affected by the same issue, well really all of them, but I'll ping you on the Artifactory PR about it specifically.

xuzhang3:
@davidcorrigan714  The hash function  is used to encode the `password/token` to forbidden this sensitive data leak to `state/console/logs`. Latest Terraform versions will forbidden the `sensitive` data output to console by default.   Encoding does not do much help.
And you are right,  the update should not be that rough, just delete does not do any compatible.  This is a breaking change to cx,  an update will be triggered after update the provider and cx just need to `apply ` it. The service connection will not be changed but update the `state`

davidcorrigan714:
@xuzhang3 Sensitive data in terraform is a very complex topic, as evident by their [Issue on it](https://github.com/hashicorp/terraform/issues/516). Outputting sensitive data to the console is one of many issues. The larger one is the handling of the data in the state file. This change effectively unmasks user data that may be stored in version control. If a user is relying on that behavior and currently storing their state files in source code this change may unmask sensitive data and cause them to inadvertently commit secrets to their VCS.

HashiCorp's [current recommendation](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/sensitive-state) seems to be that storing sensitive data in state is fine under the assumption that users are using secure storage backend. That's a bit of a controversial outlook, but users generally expect the handling of sensitive data as plain text in the state file to be explicitly called out in the provider documentation, which this change did not add and changing the behavior on existing data could very well unmask user's data who were relying on the old behavior. See the [GitHub](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_environment_secret), [AzureRM](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/key_vault_secret), or [JFrog ](https://registry.terraform.io/providers/jfrog/artifactory/latest/docs/resources/access_token) providers for examples on what users generally expect those warnings to look like.

This commit also just doesn't fix the actual issue. BCrypt should not have been used for secret memo, at all. At the least it's completely unsuitable for Generic service connections where no guarantees can be made about the token length. I'm not going to check the full list of service connection types because it's a fruitless endeavor, caring about the token format isn't something that should be handled in the provider for most cases.

xuzhang3:
@davidcorrigan714 There is a another issue needs to fixed if the sensitive data need to be encoded. Usually there is no error, but when user updates other attributes like "description". After the update, sensitive data will be set to an empty string. This is because the raw data will be set to an empty string after encoding the sensitive data. One way is always to get the data from raw configuration and ignore the state.

Issue #663, opened 2022-11-11
azuredevops_git_repository_file does not consider branch while checking for file existence


bloewes:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform 1.3.2
Azure Devops Provider 0.3.0

### Affected Resource(s)

resource azuredevops_git_repository_file

### Error description

I have repository with a default branch main. If I create resource as follows

```
   resource "azuredevops_git_repository_file" "template" {
  repository_id      = data.azuredevops_git_repository.project_repo.id
  file                      = "test.txt"
  content               = "foo"
  branch                 = "refs/heads/terraform"
  commit_message = "Update"
  overwrite_on_create = true
```
I expect the file to be created in my branch terraform. Note that this branch is empty so far.
However, my default branch main already contains a file "test.txt".
If I look at the following code line https://github.com/microsoft/terraform-provider-azuredevops/blob/270599b9a42e85380a0ac0e48ab164be922238fc/azuredevops/internal/service/git/resource_git_repository_file.go#L109-L112
and at the corresponding function GetItem, it doesn't take the branch into account, it always checks the default branch where the file already exists.

Therefore, changeType gets set to Edit not Add and the apply of the resource fails.
https://github.com/microsoft/terraform-provider-azuredevops/blob/270599b9a42e85380a0ac0e48ab164be922238fc/azuredevops/internal/service/git/resource_git_repository_file.go#L118-L124

Can you fix this call to take the branch into account correctly?

bloewes:
@xuzhang3 any chance you can have a look at this?

xuzhang3:
@bloewes  The create/update will failed if the branch not exist.  Introduce a new resource `azuredevops_repositroy_branche` to managed the branches would be the best choice.

bloewes:
Fixed by #713

PR #713, opened 2023-02-11
feat: resource_azuredevops_git_repository_branch

ivi-vink:
## All Submissions:

Issue Number: #663

Introduce the ability to manage files on branches that are managed by terraform.  Maybe an extra acceptance test case should be included that creates/updates a file on a branch managed by terraform that is not the default branch.

Basically, I took inspiration from [gitlab_branch](https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/blob/main/internal/provider/sdk/resource_gitlab_branch.go) and [github_branch](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_branch.go).

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

To reproduce bug from linked issue:
1. Create file on branch called test that is not default branch.
2. apply something like this
```hcl
data "azuredevops_project" "example" {
  name = "cdktf-test-project"
}

resource "azuredevops_git_repository" "example" {
  project_id = data.azuredevops_project.example.id
  name       = "Example Git Repository"
  initialization {
    init_type = "Clean"
  }
}

resource "azuredevops_git_repository_branch" "example" {
  repository_id = azuredevops_git_repository.example.id
  name          = "test"
}

resource "azuredevops_git_repository_file" "example" {
  repository_id       = azuredevops_git_repository.example.id
  file                = "hello"
  content             = "you"
  branch              = azuredevops_git_repository_branch.example.ref
  commit_message      = "test"
  overwrite_on_create = true
}
```

3. get logs like this

```
[mike@lemptop:~/project/fork-azdo-provider/mytests]$ terraform apply
data.azuredevops_project.example: Reading...
data.azuredevops_project.example: Read complete after 0s [id=3151f6cf-4e03-4f18-b8aa-43ffa0b1634d]
azuredevops_git_repository.example: Refreshing state... [id=d8b105e2-a69a-4ee4-ad8e-7afa5fd68d7d]
azuredevops_git_repository_branch.example: Refreshing state... [id=d8b105e2-a69a-4ee4-ad8e-7afa5fd68d7d:test]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  + create

  # azuredevops_git_repository_file.example will be created
  + resource "azuredevops_git_repository_file" "example" {
      + branch              = "refs/heads/test"
      + commit_message      = "test"
      + content             = "you"
      + file                = "hello"
      + id                  = (known after apply)
      + overwrite_on_create = true
      + repository_id       = "d8b105e2-a69a-4ee4-ad8e-7afa5fd68d7d"
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

azuredevops_git_repository_file.example: Creating...
╷
│ Error: Create repositroy file failed, repositoryID: d8b105e2-a69a-4ee4-ad8e-7afa5fd68d7d, branch: refs/heads/test, file: hello. 
Error:  The path 'hello' specified in the add operation already exists. Please specify a new path.
│ Parameter name: newPush
│ 
│   with azuredevops_git_repository_file.example,
│   on main.tf line 27, in resource "azuredevops_git_repository_file" "example":
│   27: resource "azuredevops_git_repository_file" "example" {
│ 
╵
```

Should be fixed in second commit of this pull request. Basically we check if the file exists on the right branch as suggested in the original issue and set the ChangeType accordingly.

### DIFF
--- a/azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
+++ b/azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
@@ -1,0 +1,225 @@
+//go:build (all || core || resource_git_repository_branch) && !exclude_resource_git_repository_branch
+// +build all core resource_git_repository_branch
+// +build !exclude_resource_git_repository_branch
+
+package acceptancetests
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/microsoft/azure-devops-go-api/azuredevops/v6/git"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/client"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/utils/tfhelper"
+)
+
+// TestAccGitRepoBranch_CreateUpdateDelete verifies that a branch can
+// be added to a repository and that it can be replaced
+func TestAccGitRepoBranch_CreateAndUpdate(t *testing.T) {
+	var gotBranch git.GitBranchStats
+	var gotBranch2 git.GitBranchStats
+	var gotBranch3 git.GitBranchStats
+	projectName := testutils.GenerateResourceName()
+	gitRepoName := testutils.GenerateResourceName()
+	branchName := testutils.GenerateResourceName()
+	branchNameChanged := testutils.GenerateResourceName()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testutils.PreCheck(t, nil) },
+		Providers: testutils.GetProviders(),
+		Steps: []resource.TestStep{
+			{
+				Config: hclGitRepoBranches(projectName, gitRepoName, "Uninitialized", branchName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccGitRepoBranchExists("foo_orphan", &gotBranch),
+					testAccGitRepoBranchExists("foo_from_ref", &gotBranch2),
+					testAccGitRepoBranchExists("foo_from_sha", &gotBranch3),
+					testAccGitRepoBranchAttributes("foo_orphan", &gotBranch, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch-%s", branchName),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref:        "",
+						source_sha:        false,
+						is_default_branch: true,
+						ref:               fmt.Sprintf("refs/heads/testbranch-%s", branchName),
+						sha:               true,
+					}),
+					testAccGitRepoBranchAttributes("foo_from_ref", &gotBranch2, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch2-%s", branchName),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref: fmt.Sprintf("refs/heads/testbranch-%s", branchName),
+						source_sha: true,
+						ref:        fmt.Sprintf("refs/heads/testbranch2-%s", branchName),
+						sha:        true,
+					}),
+					testAccGitRepoBranchAttributes("foo_from_sha", &gotBranch3, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch3-%s", branchName),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref: "",
+						source_sha: true,
+						ref:        fmt.Sprintf("refs/heads/testbranch3-%s", branchName),
+						sha:        true,
+					}),
+				),
+			},
+			// Test import branch created from ref
+			{
+				ResourceName:            "azuredevops_git_repository_branch.foo_from_ref",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"source_ref", "source_sha"},
+			},
+			// Test replace/update branch when name changes
+			{
+				Config: hclGitRepoBranches(projectName, gitRepoName, "Uninitialized", branchNameChanged),
+				Check: resource.ComposeTestCheckFunc(
+					testAccGitRepoBranchExists("foo_orphan", &gotBranch),
+					testAccGitRepoBranchExists("foo_from_ref", &gotBranch2),
+					testAccGitRepoBranchExists("foo_from_sha", &gotBranch3),
+					testAccGitRepoBranchAttributes("foo_orphan", &gotBranch, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch-%s", branchNameChanged),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref:        "",
+						source_sha:        false,
+						is_default_branch: true,
+						ref:               fmt.Sprintf("refs/heads/testbranch-%s", branchNameChanged),
+						sha:               true,
+					}),
+					testAccGitRepoBranchAttributes("foo_from_ref", &gotBranch2, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch2-%s", branchNameChanged),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref: fmt.Sprintf("refs/heads/testbranch-%s", branchNameChanged),
+						source_sha: true,
+						ref:        fmt.Sprintf("refs/heads/testbranch2-%s", branchNameChanged),
+						sha:        true,
+					}),
+					testAccGitRepoBranchAttributes("foo_from_sha", &gotBranch3, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch3-%s", branchNameChanged),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref: "",
+						source_sha: true,
+						ref:        fmt.Sprintf("refs/heads/testbranch3-%s", branchNameChanged),
+						sha:        true,
+					}),
+				),
+			},
+			// Test invalid ref
+			{
+				Config: fmt.Sprintf(`
+%s
+
+resource "azuredevops_git_repository_branch" "foo_nonexistent_tag" {
+	repository_id = azuredevops_git_repository.repository.id
+    name = "testbranch2-non-existent-tag"
+	source_ref = "refs/tags/non-existent"
+}
+`, hclGitRepoBranches(projectName, gitRepoName, "Clean", branchNameChanged)),
+				ExpectError: regexp.MustCompile(`No refs found that match source_ref "refs/tags/non-existent"`),
+			},
+		},
+	},
+	)
+}
+
+func testAccGitRepoBranchAttributes(node string, branch *git.GitBranchStats, want *testAccGitRepoBranchExpectedAttributes, wantState *testAccGitRepoBranchExpectedStateAttrs) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		if *branch.Name != want.Name {
+			return fmt.Errorf("Error got name %s, want %s", *branch.Name, want.Name)
+		}
+
+		rs, ok := s.RootModule().Resources[fmt.Sprintf("azuredevops_git_repository_branch.%s", node)]
+		if !ok {
+			return fmt.Errorf("Not found: %s", node)
+		}
+
+		sourceRef := rs.Primary.Attributes["source_ref"]
+		if wantState.source_ref != sourceRef {
+			return fmt.Errorf("azuredevops_git_repository_branch.%s.source_ref = %s, want %s", node, sourceRef, wantState.source_ref)
+		}
+
+		sourceSha := rs.Primary.Attributes["source_sha"]
+		if wantState.source_sha && sourceSha == "" {
+			return fmt.Errorf("azuredevops_git_repository_branch.%s.source_sha is not set", node)
+		}
+
+		isDefaultBranch := rs.Primary.Attributes["is_default_branch"]
+		if wantState.is_default_branch && isDefaultBranch != "true" {
+			return fmt.Errorf("azuredevops_git_repository_branch.%s.is_default_branch = %s, want %v", node, isDefaultBranch, wantState.is_default_branch)
+		}
+
+		ref := rs.Primary.Attributes["ref"]
+		if wantState.ref != ref {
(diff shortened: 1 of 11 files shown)


ivi-vink:

azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
```diff
@@ -0,0 +1,225 @@
...
+	branchNameChanged := testutils.GenerateResourceName()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testutils.PreCheck(t, nil) },
+		Providers: testutils.GetProviders(),
+		Steps: []resource.TestStep{
+			{
+				Config: hclGitRepoBranches(projectName, gitRepoName, "Uninitialized", branchName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccGitRepoBranchExists("foo_orphan", &gotBranch),
```
xuzhang3: Acc test will ensure the resource created.  Otherwise the test case will failed with plan not empty. Check branch exist can be removed.

azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
```diff
@@ -0,0 +1,225 @@
...
+					testAccGitRepoBranchExists("foo_from_sha", &gotBranch3),
+					testAccGitRepoBranchAttributes("foo_orphan", &gotBranch, &testAccGitRepoBranchExpectedAttributes{
+						Name: fmt.Sprintf("testbranch-%s", branchName),
+					}, &testAccGitRepoBranchExpectedStateAttrs{
+						source_ref:        "",
+						source_sha:        false,
+						is_default_branch: true,
+						ref:               fmt.Sprintf("refs/heads/testbranch-%s", branchName),
+						sha:               true,
+					}),
```
xuzhang3: Custom validation can be replace with:
```go
resource.TestCheckResourceAttr("azuredevops_git_repository_branch.foo_orphan", "name", fmt.Sprintf("testbranch-%s", branchName)),
resource.TestCheckResourceAttr("azuredevops_git_repository_branch.foo_orphan", "is_default_branch", "true"),
resource.TestCheckResourceAttr("azuredevops_git_repository_branch.foo_orphan", "ref", fmt.Sprintf("refs/heads/testbranch-%s", branchName)),
resource.TestCheckResourceAttrSet("azuredevops_git_repository_branch.foo_orphan", "sha"),

```

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,285 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGitRepositoryBranchImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.NoZeroValues,
+				Sensitive:    false,
```
xuzhang3: can be deleted as default is `Sensitive:false`

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+func resourceGitRepositoryBranchImport(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+	_, branchName, err := tfhelper.ParseGitRepoBranchID(d.Id())
+	if err != nil {
+		return nil, err
+	}
+
+	diags := resourceGitRepositoryBranchRead(ctx, d, m)
```
xuzhang3: no need to call the `read` in the import. Make sure to set an ID for this resource and Terraform will automatically call `read`.

xuzhang3 (changes requested):
@mvinkio Can we change the `source_ref` to `ref`, `tag` or `commitId` like the UI pages? By skip the `source_ref`, a totally new branch without any other commit history will be created..

ivi-vink:
> @mvinkio Can we change the `source_ref` to `ref`, `tag` or `commitId` like the UI pages? 

Ok, the resource schema should indeed have ref, tag, and commitId fields like in the azure devops GUI when creating new branch.

> By skip the `source_ref`, a totally new branch without any other commit history will be created..

This creation of branch without existing commit history will be removed.

azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
ivi-vink: Removed and replace checks with TestCheckResourceAttr.

azuredevops/internal/acceptancetests/resource_git_repository_branch_test.go
ivi-vink: Replaced checks in acceptance tests.

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Mistake removed.

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Removed unecessary call to read method.

ivi-vink:
Also acceptance test TestAccGitRepoFile_Create_IncorrectBranch fails on my machine, it tries to apply regex too the error from the azure devops sdk GetBranch.

https://github.com/mvinkio/terraform-provider-azuredevops/blob/d3913879bbf8ee1224afe25637c3de1afef551f3/azuredevops/internal/service/git/resource_git_repository_file.go#L106-L108

It works again if I change the regex pattern to `Error: Branch "foobar" does not exist in the \S+ repository.`. Not sure what causes the different error message. Maybe I'm using wrong terraform version or something?

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,239 @@
...
+				ForceNew:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
+			"commit_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				Computed:     true,
+				ValidateFunc: validation.StringIsNotEmpty,
+			},
```
xuzhang3: Suggest rename  `ref` -> `ref_branch`,  `tag` -> `ref_tag` , `commit_id` -> `ref_commit_id` and only one of them can be configured at once. `ConflictsWith` can be used to add such constraint.
```suggestion
"ref_branch": {
	Type:          schema.TypeString,
	Optional:      true,
	ForceNew:      true,
	ValidateFunc:  validation.StringIsNotEmpty,
	ConflictsWith: []string{"ref_tag", "ref_commit_id"},
},
"ref_tag": {
	Type:          schema.TypeString,
	Optional:      true,
	ForceNew:      true,
	ValidateFunc:  validation.StringIsNotEmpty,
	ConflictsWith: []string{"ref_tag", "ref_commit_id"},
},
"ref_commit_id": {
	Type:          schema.TypeString,
	Optional:      true,
	ForceNew:      true,
	Computed:      true,
	ValidateFunc:  validation.StringIsNotEmpty,
	ConflictsWith: []string{"ref_tag", "ref_commit_id"},
},
```

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,239 @@
...
+			"branch_reference": {
```
xuzhang3: `branch_reference` is bit duplicate of `name` , the only difference is the prefix

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,239 @@
...
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"branch_head": {
```
xuzhang3: what is this property for? It sets the `commit ID` in `read`, more meaningful name should be `last_commit_id`?

xuzhang3 (changes requested):

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Ok, I think this is good indeed.

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: I think `name` should be without prefix, because GetBranch api expects `name` without prefix and in the GUI pages it is also without prefix.

But other resources in the azure devops provider like git_repository_file, git_repository, branch policies expect branch name with prefix. This `branch_reference` field can be used to pass valid branch name to those.

 Some options are:
- Remove this field and force name to have prefix
- Remove this field and keep name the same without prefix
- Remove this field and change git_repository_file and other resources to use branch name without prefix

My preference is the first or the second one. The first one is  more like branch names in other resources.  In the second one we can let users add prefix like `"refs/heads/${azuredevops_git_repository_branch.example.name}"`.

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Ok, this is better name indeed. Branch head is a file in git repository with last commit id.

azuredevops/internal/service/git/resource_git_repository_branch.go
mvinkio: Can also remove `branch_reference` and allow `name` both with and without prefix.

ivi-vink: For now I went with option 2. But can easily change to option 1. Curious what others think.

xuzhang3: I prefer the second one, let the user decide which prefix they want.

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,265 @@
...
+				Optional:      true,
+				ForceNew:      true,
+				Computed:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				ConflictsWith: []string{"ref_branch", "ref_tag"},
+			},
+			"last_commit_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
xuzhang3: `ref_commit_id` should not be an `computed`, latest commit ID of `ref_branch` will change over development and the `ref_commit_id` will be updated as well.
My understanding of this API is `ref_branch`, `ref_commit_id` and `ref_commit_id` are conflict with each other and at least one and only one of them needs to be configured in the HCL. They need to add a `ExactlyOneOf:  []string{"ref_branch", "ref_tag", "ref_commit_id"},` constraint.

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,265 @@
...
+		if gotRef.PeeledObjectId != nil {
+			refObjectIdSha = gotRef.PeeledObjectId
+		} else if gotRef.ObjectId != nil {
+			refObjectIdSha = gotRef.ObjectId
+		} else {
+			return diag.FromErr(fmt.Errorf("GetRefs response doesn't have a valid commit id."))
+		}
+		d.Set("ref_commit_id", *refObjectIdSha)
+	}
+	newObjectId := d.Get("ref_commit_id").(string)
```
xuzhang3:
```suggestion
	var newObjectId string
	if v, ok := d.GetOk("ref_commit_id"); ok {
		newObjectId = v.(string)
	} else {
		var rs string
		if v, ok := d.GetOk("ref_branch"); ok {
			rs = withPrefix(REF_BRANCH_PREFIX, v.(string))
		}
		if v, ok := d.GetOk("ref_tag"); ok {
			rs = withPrefix(REF_TAG_PREFIX, v.(string))
		}

		filter := strings.TrimPrefix(rs, "refs/")
		gotRefs, err := clients.GitReposClient.GetRefs(clients.Ctx, git.GetRefsArgs{
			RepositoryId: converter.String(repoId),
			Filter:       converter.String(filter),
			Top:          converter.Int(1),
			PeelTags:     converter.Bool(true),
		})
		if err != nil {
			return diag.FromErr(fmt.Errorf("Error getting refs matching %q: %w", filter, err))
		}

		if len(gotRefs.Value) == 0 {
			return diag.FromErr(fmt.Errorf("No refs found that match ref %q.", rs))
		}

		gotRef := gotRefs.Value[0]
		if gotRef.Name == nil {
			return diag.FromErr(fmt.Errorf("Got unexpected GetRefs response, a ref without a name was returned."))
		}

		// Check for complete match. Sometimes refs exist that match prefix with Ref, but do not match completely.
		if *gotRef.Name != rs {
			return diag.FromErr(fmt.Errorf("Ref %q not found, closest match is %q.", filter, *gotRef.Name))
		}

		if gotRef.PeeledObjectId != nil {
			newObjectId = *gotRef.PeeledObjectId
		} else if gotRef.ObjectId != nil {
			newObjectId = *gotRef.ObjectId
		} else {
			return diag.FromErr(fmt.Errorf("GetRefs response doesn't have a valid commit id."))
		}
	}

```

azuredevops/internal/service/git/resource_git_repository_branch.go
```diff
@@ -0,0 +1,265 @@
...
+	REF_TAG_PREFIX    = "refs/tags/"
+)
+
+// ResourceGitRepositoryBranch schema to manage the lifecycle of a git repository branch
+func ResourceGitRepositoryBranch() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGitRepositoryBranchCreate,
+		ReadContext:   resourceGitRepositoryBranchRead,
+		DeleteContext: resourceGitRepositoryBranchDelete,
+		Importer: &schema.ResourceImporter{
```
xuzhang3: Get refs API only have the object ID, import this resource may always force a new resource to be created, Recommended that this resource does not support importing.

xuzhang3 (changes requested):

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Yea, I kind of looked at what they did in github's implementation. They only set `ref_commit_id` during creation given by user or by the other ref. It indeed creates a double and confusing meaning for `ref_commit_id`. I think it's indeed fine to remove this computed setting part for now.

azuredevops/internal/service/git/resource_git_repository_branch.go
ivi-vink: Probably we can add better import that also supports setting `ref_tag`, `ref_branch` or `ref_commit_id` manually using import expression. Something like this is also done in `github_branch`:

`<repository_id>:<name>:<ref_(commit_id | tag | branch)>`

Not sure if necessary. I'll remove import for now I think and maybe make another issue.

xuzhang3 (approved):

xuzhang3:
@mvinkio  LGTM 🎆

PR #747, opened 2023-04-06
Service Principal Authentication

### LINKED ISSUE #745: Service Principal Authentication Support
### Community Note

### Description

The provider should support authenticating as a Service Principal to Azure DevOps.

```hcl
provider "azuredevops" {
  org_service_url = "https://dev.azure.com/my-org"
  sp_client_id    = "00000000-0000-0000-0000-000000000001"
  sp_tenant_id    = "00000000-0000-0000-0000-000000000001"

  sp_client_secret = "a client secret from AzureAD"

  # sp_client_secret_path = "C:\\client_secret.txt"

  # sp_oidc_token = "my long oidc token"
  # sp_oidc_token_path = "C:\\oidc_token.txt"
  # sp_oidc_github_actions = true
  # sp_oidc_github_actions_audience = "my-aud"
  # sp_oidc_hcp = true
  # sp_client_certificate_path = "C:\\cert_tes.pfx"
  # sp_client_certificate          = "base64 encoded cert"
  # sp_client_certificate_password = "cert password"
}

```

### References
[Use Azure AD service principals & managed identities](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?toc=%2Fazure%2Fdevops%2Forganizations%2Fsecurity%2Ftoc.json&view=azure-devops)

davidcorrigan714:
## All Submissions:

Allows the terraform provider to authenticate to azure devops through a service principal using a variety of authentication methods: client secret, client certificate, GitHub Actions OIDC tokens, Terraform Cloud OIDC tokens, & any other OIDC token provided directly as an attribute

Issue Number: #745 

- [x] Yes
- [ ] No

Adds the Azure Go SDK & some crypto functions.

- [ ] Yes
- [x] No


website/docs/index.html.markdown
```diff
@@ -31,13 +31,84 @@ resource "azuredevops_project" "project" {
 }
 ```
 
+## Authentication
+
+Authentication may be accomplished using an [Azure AD service principal](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity) if your organization is coonnected to Azure AD,
```
cveld: typo: coonnected -> connected

xuzhang3:
@davidcorrigan714 can you help rebase to the latest code, we upgrade the SDK from v6 to v7,.

davidcorrigan714:
Looks like the project has been busy since last I was able to check things. Got an initial round of cleanup on this, might be good enough for an initial review. I'm not convinced the vendoring & mod files are quite right at the moment and I need to rerun the tests which I can do tomorrow.

davidcorrigan714:
Fixed up some test issues, and did a manual test on my machine with just the client secret auth method. I tried to do thorough unit tests, but the acceptance tests are a bit lacking just because it would depend on having a service principal configured for the test environment.

azuredevops/provider.go
```diff
@@ -141,6 +167,134 @@ func Provider() *schema.Provider {
 				Description: "The personal access token which should be used.",
 				Sensitive:   true,
 			},
+			"sp_client_id": {
```
jaredfholgate: @davidcorrigan714 Apologies if this is already documented somewhere, but I'm wondering why the choice was made to use different env vars to the `azurerm`, `azuread` and `azapi` providers? If it worked the same as those providers by default, it might reduce the setup complexity for some users. E.g. `ARM_USE_OIDC`, etc.

azuredevops/provider.go
```diff
@@ -158,7 +547,12 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
 			terraformVersion = "0.11+compatible"
 		}
 
-		client, err := client.GetAzdoClient(d.Get("personal_access_token").(string), d.Get("org_service_url").(string), terraformVersion)
+		token, err := GetAuthToken(ctx, d, AzIdentityFuncsReal{})
+		if err != nil {
+			return nil, diag.FromErr(err)
+		}
+
+		client, err := client.GetAzdoClient(token, d.Get("org_service_url").(string), terraformVersion)
```
tmeckel: The call to `GetAzdoClient` will call `NewPatConnection`, what in turn will create a Basic Auth HTTP header using an OAuth token, what eventually will fail the authentication, because OAuth tokens must be passed as a `Bearer`  Auth Header.

An even bigger issue is here that the Azure DevOps GO API uses a `struct` instead of an `interface` to pass a `Connection` around to the various REST API endpoint implementations and the `Connection` struct only supports a static auth header value. This is sufficient for a long living PAT but far from ideal for short living OAuth tokens. As long as this issue isn't fixed in the GO API is don't see a reliable way to support Service Principal Authentication for Azure DevOps.

@xuzhang3 any infos who's actually in charge for the GO API? Or is it abandoned again?

https://github.com/microsoft/terraform-provider-azuredevops/issues/745#issuecomment-1717289248

azuredevops/provider.go
davidcorrigan714: I can go double check this, the main concern is ensuring that they can be configured independently when needed and support multiple instances to different accounts. Not sure the Azure one suffered that bug but the AWS implementation had issues with that, I should find that bug and double check what they did there. The environment variable names for them do seem to provider specific though, like this isn't Azure Resource manager so ARM_.... doesn't really make sense to me

azuredevops/provider.go
davidcorrigan714: That's a good point, I thought the OAuth tokens were valid long enough that it wouldn't practically be a concern. Looking at your other comment on the other thread I can test that and put it in a PR up to the Go SDK if someone is available to review it.

tmeckel: @davidcorrigan714 The issue with the Azure DevOps GO SDK ist, that it's generated using a framework not created manually. So that's why most PRs got rejected there in past, because issues have to be fixed in the framework (which is not available here on GitHub) and not in the generated code in the repo you see. Of course, the fact that the AzDO team maintains the GO SDK so carelessly makes things even worse. This is one of the main reasons why I personally stopped contributing to the project and in addition to me I think Microsoft focuses on GitHub and slowly let AzDO die. That AzDO has much more stability issues compared to GitHub is a another good indicator that the manpower behind AzDO is very limited.

davidcorrigan714: Oh I'm well aware of the code-gen happening in all the SDKs. Pretty sure they completely lost or broke the Python SDK generator tooling a few years back, or lost the expertise that was running it. connection.go just doesn't show any history of ever being a generated file like the rest of the SDK is.

Don't want to derail this into GH vs AzDO too much. The reality is they don't have feature parity on the roadmap for the two platforms in certain areas and the rate they've been improving since trying to pitch GH as an AzDO alternative still puts it at least 5+ years out as a viable alternative for most of our use cases. We also do some "interesting" things with Pipelines that GH may never support in Actions.

xuzhang3:
@davidcorrigan714  can you help resolve the conflict?

azuredevops/provider.go
cveld: Any update? I would say ship a first version without refresh token support 😊

jessehouwing: The Azure DevOps team doesn't really have go developers, Azure DevOps is built in .NET, PowerShell and a bit of NodeJS and some web front-end stuff. The fact they even have a Go SDK amazes me, the team generates the Go and the Node SDK from the .NET API metadata as far as I know.

I don't think they maintain these libraries "carelessly", it's just that the person who wrote it, probably isn't available at the moment.

It's one of the big pains of building SDKs at all, you tend to need to build them in a plethora of languages, Python, .NET, JS, Go, Ruby etc, but the team that builds the SDK doesn't normally work with any of thee languages other than .NET. I see the same issue with many other frameworks and SDKs, the best SDK is written in for the language of the tool itself.

geekzter: The team maintaining TF az* providers is a central team. 
While you wait, consider [this workaround](https://github.com/geekzter/azure-identity-scripts/blob/123e9764195c8661e28a7dc189b2691447dad4bb/terraform/azure-devops/create-service-connection/terraform.tf#L27) that uses the `external` datasource to acquire a token with the Azure CLI:

```hcl
data external azdo_token {
  program                      = [
    "az", "account", "get-access-token", 
    "--resource", "499b84ac-1321-427f-aa17-267ca6975798", # Azure DevOps
    "--query","{accessToken:accessToken}",
    "-o","json"
  ]
}
provider azuredevops {
  org_service_url              = local.azdo_organization_url
  personal_access_token        = data.external.azdo_token.result.accessToken
}
```

davidcorrigan714: @jessehouwing yeah the choice of Go for Terraform always makes me a bit sad, it's the only reason I've picked up Go and I'm certainly not amazing at it though the Terraform stuff usually has plenty of examples to draw from.

Catching up on this thread again now and seeing where things are at with the code.

XtratusCloud:
I am very interested in the final implementation of this functionality, as it would enable the use of this functionality for authentication through Managed Identities, on the road to password-less processes.
Please, if it would be possible to complete this pull request to include this functionality in the next version? 
It would be of great benefit.
Thanks,

davidcorrigan714:
Got most of the feedback incorporated with refreshing the access token as needed and aligning with the other Azure terraform provider variables where possible. Going to do a bit of manual testing tomorrow for the various cloud identity scenarios - GH, Managed Identities & HCP Workload Identities. The changes to the Azure DevOps Go API are currently hacked into it in the vendor folder so I need to do a PR there before this goes in.

davidcorrigan714:
Maybe @xuzhang3  can provide some insights on what's failing in the automated checks. Wouldn't surprise me if I missed a format or package checks but can't see the logs ☹

xuzhang3:
> Maybe @xuzhang3 can provide some insights on what's failing in the automated checks. Wouldn't surprise me if I missed a format or package checks but can't see the logs ☹

@davidcorrigan714 The pipe fails due to the line break in the vendor files.  I usually add the git config ` git config --global core.autocrlf true`  to my local env to ensure that different end-of-file lines will be managed automatically by git. Since the vendor files has been add pushed to this branch you need ` git config --global core.autocrlf false` and push the files again

valentin-fischer:
Can anyone check and fix the build so we can get this merged? Thanks!

troyel:
@davidcorrigan714 Looks to me like you have a lot of great work in here! Do you think you have any time to take this through the home stretch? I think this would be very useful for a lot of people and orgs! 🙌

website/docs/guides/authenticating_managed_identity.html.md
```diff
@@ -0,0 +1,60 @@
...
+### Configuring with environment variables
+
+The `use_msi` must be set to `true` to use a managed identity. By default, Terraform will use the system assigned identity for authentication. To use a user assigned identity instead, you will need to specify the `ARM_CLIENT_ID` environment variable (equivalent to provider block argument [`client_id`](https://registry.terraform.io/providers/azure/azapi/latest/docs#client_id)) to the [client id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/user_assigned_identity#client_id) of the identity.
+
+A provider block is _technically_ optional when using environment variables. Even so, we recommend defining provider blocks so that you can pin or constrain the version of the provider being used, and configure other optional settings:
+
+```hcl
+terraform {
+  required_providers {
+    azapi = {
```
xuzhang3, line 32: should be `azuredevops` not `azapi`

website/docs/guides/authenticating_managed_identity.html.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+### Configuring with the provider block
+
+It's also possible to configure a managed identity within the provider block:
+
+```hcl
+terraform {
+  required_providers {
+    azapi = {
```
xuzhang3, line 50: should be `azuredevops` not `azapi`

website/docs/guides/authenticating_service_principal_using_an_oidc_token.html.md
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+}
+
+provider "azuredevops" {
+  org_service_url                 = "https://dev.azure.com/my-org"
+
+  client_id = "00000000-0000-0000-0000-000000000001"
+  tenant_id = "00000000-0000-0000-0000-000000000001"
+  oidc_hcp  = true
```
xuzhang3, line 108: `oidc_hcp`  is not supported

website/docs/guides/authenticating_service_principal_using_an_oidc_token.html.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+provider "azuredevops" {
+  org_service_url                 = "https://dev.azure.com/my-org"
+
+  client_id_plan  = "00000000-0000-0000-0000-000000000001"
+  client_id_apply = "00000000-0000-0000-0000-000000000001"
+  tenant_id_plan  = "00000000-0000-0000-0000-000000000001"
+  tenant_id_apply = "00000000-0000-0000-0000-000000000001"
+  oidc_hcp  = true
```
xuzhang3, line 136: `oidc_hcp`  is not supported

website/docs/index.html.markdown
```diff
@@ -31,13 +31,99 @@ resource "azuredevops_project" "project" {
 }
 ```
 
+## Authentication
+
+Authentication may be accomplished using an [Azure AD service principal](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity) if your organization is connected to Azure AD, or by a [personal access token](https://learn.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate).
+
+The OIDC service principal authentication methods allow for secure passwordless authentication from [Terraform Cloud](https://developer.hashicorp.com/terraform/cloud-docs/workspaces/dynamic-provider-credentials) & [GitHub Actions](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect).
+
+* [Authenticating to a Service Principal with a Terraform Cloud Workload Identity Token](guides/authenticating_service_principal_using_hcp_token.html)
```
xuzhang3, line 40: Documention `authenticating_service_principal_using_hcp_token.html`  does not exist，should be the manage identity document

website/docs/index.html.markdown
```diff
@@ -31,13 +31,99 @@ resource "azuredevops_project" "project" {
...
+- `oidc_token_file_path` - The path to a file containing nn OIDC token to authenticate to a service principal.
+It can also be sourced from the `AZDO_TOKEN_PATH` environment variable.
+
+- `oidc_github_actions` - Boolean, set to true to use a GitHub Actions OIDC token to authenticate to a service principal.
+It can also be sourced from the `ARM_OIDC_GITHUB_ACTIONS` environment variable.
+
+- `oidc_github_actions_audience` - Custom audience for the GitHub Actions OIDC token.
+It can also be sourced from the `ARM_OIDC_GITHUB_ACTIONS_AUDIENCE` environment variable.
+
+- `oidc_hcp` - Boolean, set to true to use the Terraform Cloud OIDC workload identity token to authenticate to a service principal.
```
xuzhang3, line 113: `oidc_hcp` does not exist

xuzhang3 (changes requested):

xuzhang3:
@davidcorrigan714 if you do not have time to update this PR, can we create another PR based on your changes?

xuzhang3:
@davidcorrigan714  close this PR via #970

azuredevops/provider.go
jubr: @davidcorrigan714 @jaredfholgate @xuzhang3 whatever happened to not using `ARM_*` vars? This is now breaking pipelines that have (a subset) of those vars set for the other `azure*` providers, but have no relevance for `azuredevops`.

I understand the — let's make this easy for people so it "just works" — sentiment, but in practice this will "just break" things for people in an unsuspecting way if they (or their pipeline) use a combination of the `azure{rm,ad,devops}` providers. Which will not be an empty groups me thinks. $0.02 given.

Thinking through on this, how about when this (for some people unintended) OIDC authentication is attempted and it fails, the provider shows a hint in how this new behavior can be disabled. Perhaps with ~the code snippet I give in the comment below~, or a more elegant way like for instance `ignore_arm_env_vars = true`. #OpenForExtensionClosedForModification

jubr: Here's a workaround in case someone else is stumbling onto this:
```terraform
provider "azuredevops" {
  // … other settings

  // prevent enabling OIDC when `ARM_*` vars are detected in CI
  use_oidc  = false
  // not used, but these need to be set to valid GUID if `use_oidc` is set
  client_id = "deadbeef-dead-d34d-dead-deadbeefdead"
  tenant_id = "deadbeef-dead-d34d-dead-deadbeefdead"
}
```
**_Edit_**: turns out this does not work 😢 after all:
```terraform
│ Error: Invalid combination of arguments
│ 
│   with provider["registry.terraform.io/microsoft/azuredevops"],
│   on _providers.tf line 21, in provider "azuredevops":
│   21: provider "azuredevops" {
│ 
│ "client_secret_path": only one of
│ `client_certificate,client_certificate_path,client_secret,client_secret_path,personal_access_token,use_msi,use_oidc`
│ can be specified, but `personal_access_token,use_oidc` were specified.
```

Pinning to `version = "~> 0.11"` until this is resolved or until we have time to refactor our authentication configuration.

davidcorrigan714:
@xuzhang3 appreciate you all finishing this up and getting it in. Life has been too darn busy and still is.

PR #806, opened 2023-07-01
New resource - `azuredevops_check_required_template`

rdalbuquerque:
## All Submissions:

Adding new resource `azuredevops_check_required_template` to manage required template check.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
+++ b/azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
@@ -8,7 +8,7 @@ import (
 	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/utils/pipelineschecksextras"
 )
 
-var validRepositoryTypes = []string{"git", "github", "bitbucket"}
+var validRepositoryTypes = []string{"azuregit", "github", "bitbucket"}
 
 // ResourceCheckRequiredTemplate schema and implementation for required template check resources
 func ResourceCheckRequiredTemplate() *schema.Resource {
@@ -22,7 +22,7 @@ func ResourceCheckRequiredTemplate() *schema.Resource {
 				"repository_type": {
 					Type:         schema.TypeString,
 					Optional:     true,
-					Default:      "git",
+					Default:      "azuregit",
 					ValidateFunc: validation.StringInSlice(validRepositoryTypes, false),
 				},
 				"repository_name": {
@@ -37,7 +37,7 @@ func ResourceCheckRequiredTemplate() *schema.Resource {
 				},
 				"template_path": {
 					Type:         schema.TypeString,
-					Optional:     true,
+					Required:     true,
 					ValidateFunc: validation.StringIsNotEmpty,
 				},
 			},
@@ -68,8 +68,14 @@ func flattenCheckRequiredTemplate(d *schema.ResourceData, check *pipelineschecks
 		extendsChecks := extendsChecksConfig.([]interface{})
 		for _, ec := range extendsChecks {
 			ecMap := ec.(map[string]interface{})
+			var repositoryType string
+			if ecMap["repositoryType"].(string) == "git" {
+				repositoryType = "azuregit"
+			} else {
+				repositoryType = ecMap["repositoryType"].(string)
+			}
 			reqTempl := map[string]interface{}{
-				"repository_type": ecMap["repositoryType"],
+				"repository_type": repositoryType,
 				"repository_name": ecMap["repositoryName"],
 				"repository_ref":  ecMap["repositoryRef"],
 				"template_path":   ecMap["templatePath"],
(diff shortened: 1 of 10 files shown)


rdalbuquerque:

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/utils/pipelineschecksextras"
+)
+
+var validRepositoryTypes = []string{"git", "github", "bitbucket"}
+
+// ResourceCheckRequiredTemplate schema and implementation for required template check resources
+func ResourceCheckRequiredTemplate() *schema.Resource {
+	r := genBaseCheckResource(flattenCheckRequiredTemplate, expandCheckRequiredTemplate)
+
+	r.Schema["required_template"] = &schema.Schema{
```
xuzhang3, line 17: flatten `repository_type` , `repository_name`  etc.  to the root schema without nesting block `required_template`

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/utils/pipelineschecksextras"
+)
+
+var validRepositoryTypes = []string{"git", "github", "bitbucket"}
```
xuzhang3: Can we rename `git` to `azuregit`? `git` and `github` will confuse customers.

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
```diff
@@ -0,0 +1,104 @@
...
+					Type:         schema.TypeString,
+					Required:     true,
+					ValidateFunc: validation.StringIsNotEmpty,
+				},
+				"repository_ref": {
+					Type:         schema.TypeString,
+					Required:     true,
+					ValidateFunc: validation.StringIsNotEmpty,
+				},
+				"template_path": {
```
xuzhang3, line 39: `template_path` should be `Required: true` not `Optional: true,`

xuzhang3 (changes requested):

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
rdalbuquerque: The idea of the nested block was to allow users to specify more than one required template, following the api contract.

I put an example in the docs:
```hcl
resource "azuredevops_check_required_template" "example" {
  project_id           = azuredevops_project.example.id
  target_resource_id   = azuredevops_environment.example.id
  target_resource_type = "environment"

  required_template {
    repository_name = "project/repository"
    repository_ref  = "refs/heads/main"
    template_path   = "template/path.yml"
  }

  required_template {
    repository_name = "project/repository"
    repository_ref  = "refs/heads/main"
    template_path   = "template/alternate-path.yml"
  }
}
```

I guess for most use cases only one required template will be specified but since the api allowed for more than one I went with it.

But if you think it's better design to keep everything in the root schema and allow for only 1 required template to be set please let me know. :)

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
rdalbuquerque: fixed

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
rdalbuquerque: done

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
xuzhang3: 1. Add  a `MinItems: 1,` constraint to ensure at lease one  template configured. 
2. Since all the properties within `required_template` can be updated, suggest  change the type from `TypeSet` to `TypeList`  if service will not change the request template order. Because type set will generate the index based on the hash code, as a result any change of the `required_template` block will like:
  image

xuzhang3 (changes requested):

azuredevops/internal/service/approvalsandchecks/resource_check_required_template.go
rdalbuquerque: Done.

One scenario that might be confusing is the following:
1. create resource with 2 `required_template` blocks: 
```hcl
resource "azuredevops_check_required_template" "example_vg" {
  project_id           = azuredevops_project.example.id
  target_resource_id   = azuredevops_variable_group.example.id
  target_resource_type = "variablegroup"
  required_template {
    repository_type = "azuregit"
    repository_name = "templates"
    repository_ref  = "refs/heads/master"
    template_path   = "templates/azure-pipelines.yml"
  }
  required_template {
    repository_type = "azuregit"
    repository_name = "templates"
    repository_ref  = "refs/heads/master"
    template_path   = "templates/alternate-azure-pipelines.yml"
  }
}
```
2. delete the first one from the list:
```hcl
resource "azuredevops_check_required_template" "example_vg" {
  project_id           = azuredevops_project.example.id
  target_resource_id   = azuredevops_variable_group.example.id
  target_resource_type = "variablegroup"
  required_template {
    repository_type = "azuregit"
    repository_name = "templates"
    repository_ref  = "refs/heads/master"
    template_path   = "templates/alternate-azure-pipelines.yml"
  }
}
```
as a result the plan will be like:
image

as if you were deleting the second one and changing the first one. But I guess that's expected.

xuzhang3: Usually this kind of constraint  relies strongly on the configuration order,   update the template without changing the order is acceptable.

xuzhang3 (approved):

xuzhang3:
@rdalbuquerque  LGTM

Issue #303, opened 2021-02-15
typos in error message


FernandoMiguel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform v0.14.6
+ provider registry.terraform.io/hashicorp/azuread v1.3.0
+ provider registry.terraform.io/hashicorp/azurerm v2.46.1
+ provider registry.terraform.io/hashicorp/random v3.0.1
+ provider registry.terraform.io/hashicorp/time v0.6.0
+ provider registry.terraform.io/microsoft/azuredevops v0.1.2

### Affected Resource(s)

* `azuredevops_variable_group`

### Debug Output
Error: Expanding variable group resource data: Failed to get the Azure Key **valut** secrets.Code: forbidden, error **messge**:

FernandoMiguel:
https://github.com/microsoft/terraform-provider-azuredevops/blob/33b33401b4539cd9ef245155e3cb8ca5fdd247ed/azuredevops/internal/service/taskagent/resource_variable_group.go#L429
https://github.com/microsoft/terraform-provider-azuredevops/blob/33b33401b4539cd9ef245155e3cb8ca5fdd247ed/azuredevops/internal/service/taskagent/resource_variable_group.go#L451
and other places

xuzhang3:
Hi @FernandoMiguel   This is an open source project, you can submit a PR to  fix this issues.

PR #826, opened 2023-07-11
Fixing bug where variable group fails to map new variable after service endpoint secret rotates

### LINKED ISSUE #825: `azuredevops_variable_group` key vault expansion error after azurerm service endpoint secret rotation
### Community Note

terraform version: 1.5.1
provider: 0.6.0

### Affected Resource(s)

* `azuredevops_variable_group`

```hcl
resource "azuread_application" "this" {
  display_name = "azdo-conn"
}

resource "azuread_service_principal" "this" {
  application_id = azuread_application.this.application_id
}

resource "azuread_application_password" "this" {
  application_object_id = azuread_application.this.object_id
}

resource "azuredevops_serviceendpoint_azurerm" "example" {
  project_id            = azuredevops_project.example.id
  service_endpoint_name = "Example AzureRM"
  description           = "Managed by Terraform"
  credentials {
    serviceprincipalid  = azuread_service_principal.this.application_id
    serviceprincipalkey = azuread_application_password.this.value
  }
  azurerm_spn_tenantid      = data.azuread_client_config.current.tenant_id
  azurerm_subscription_id   = data.azurerm_client_config.current.subscription_id
  azurerm_subscription_name = "mysubscriptionname"
}

resource "azurerm_key_vault" "this" {
  name                      = "example-kv1"
  resource_group_name       = azurerm_resource_group.this.name
  enable_rbac_authorization = false
  location                  = "eastus"
  sku_name                  = "standard"
  tenant_id                 = data.azurerm_client_config.current.tenant_id
  access_policy {
    tenant_id          = data.azurerm_client_config.current.tenant_id
    object_id          = data.azurerm_client_config.current.object_id
    secret_permissions = ["Get", "List", "Set", "Delete", "Purge"]
  }
  access_policy {
    tenant_id          = data.azurerm_client_config.current.tenant_id
    object_id          = azuread_service_principal.this.object_id
    secret_permissions = ["Get", "List"]
  }
}

locals {
  vars = [
    "key-1",
    "key-2" # added in the same apply as service endpoint secret rotation
  ]
}

resource "azurerm_key_vault_secret" "var" {
  for_each     = toset(local.vars)
  key_vault_id = azurerm_key_vault.this.id
  name         = each.key
  value        = "value-of-${each.key}"
}

resource "azuredevops_variable_group" "example" {
  project_id   = azuredevops_project.example.id
  name         = "Example Variable Group"
  description  = "Example Variable Group Description"
  allow_access = true

  key_vault {
    name                = azurerm_key_vault.this.name
    service_endpoint_id = azuredevops_serviceendpoint_azurerm.example.id
  }

  dynamic "variable" {
    for_each = toset(local.vars)
    content {
      name = variable.key
    }
  }
}
```

### Expected Behavior

New variable mapped to variable group after service endpoint was updated

### Actual Behavior

Variable group update fails with error `AADSTS7000215: Invalid client secret provided.`

image

### Steps to Reproduce

With the code above and the other omitted resources declared
1. Add new variable to `local.vars` list
2. run `terraform apply -replace azuread_application_password.this -replace azuread_application_password.this` to rotate secret
3. run `terraform apply` without changing anything

on step 2 the provider fails to map new variable with the error mentioned above
on step 3 the new variable is mapped to the variable group successfully
So it looks like it's taking a few seconds for Azure to propagate the new secret 
### References

* #0000

rdalbuquerque:
## All Submissions:

Adding a key vault expansion retry logic in case there is an update to azurerm service endpoint secret.

Issue Number: [#825](https://github.com/microsoft/terraform-provider-azuredevops/issues/825)

Also fixing error messages typos
Issue Number: [#303](https://github.com/microsoft/terraform-provider-azuredevops/issues/303)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

image

## Other information
The tradeoff is that if the secret is in fact wrong it will take 30 seconds to return an error

### DIFF
--- a/azuredevops/internal/acceptancetests/resource_serviceendpoint_azurecr_test.go
+++ b/azuredevops/internal/acceptancetests/resource_serviceendpoint_azurecr_test.go
@@ -14,7 +14,6 @@ import (
 // validates that an apply followed by another apply (i.e., resource update) will be reflected in AzDO and the
 // underlying terraform state.
 func TestAccServiceEndpointAzureCR_CreateAndUpdate(t *testing.T) {
-	t.Skip("Skipping test TestAccServiceEndpointAzureCR_CreateAndUpdate: test resource limit")
 	projectName := testutils.GenerateResourceName()
 	serviceEndpointNameFirst := testutils.GenerateResourceName()
 	serviceEndpointNameSecond := testutils.GenerateResourceName()
@@ -22,7 +21,9 @@ func TestAccServiceEndpointAzureCR_CreateAndUpdate(t *testing.T) {
 	resourceType := "azuredevops_serviceendpoint_azurecr"
 	tfSvcEpNode := resourceType + ".serviceendpoint"
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		PreCheck: func() {
+			testutils.PreCheck(t, &[]string{"TEST_ARM_SUBSCRIPTION_ID", "TEST_ARM_SUBSCRIPTION_NAME", "TEST_ARM_TENANT_ID"})
+		},
 		Providers:    testutils.GetProviders(),
 		CheckDestroy: testutils.CheckServiceEndpointDestroyed(resourceType),
 		Steps: []resource.TestStep{
--- a/azuredevops/internal/service/serviceendpoint/commons.go
+++ b/azuredevops/internal/service/serviceendpoint/commons.go
@@ -3,7 +3,6 @@ package serviceendpoint
 import (
 	"fmt"
 	"log"
-	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -14,15 +13,11 @@ import (
 	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/client"
 	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/utils"
 	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/utils/converter"
-	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/utils/tfhelper"
 )
 
 const errMsgTfConfigRead = "Error reading terraform configuration: %+v"
 const errMsgServiceCreate = "Error looking up service endpoint given ID (%s) and project ID (%s): %v "
 const errMsgServiceDelete = "Error delete service endpoint. ServiceEndpointID: %s, projectID: %s. %v "
-
-type flatFunc func(d *schema.ResourceData, serviceEndpoint *serviceendpoint.ServiceEndpoint, projectID *uuid.UUID)
-type expandFunc func(d *schema.ResourceData) (*serviceendpoint.ServiceEndpoint, *uuid.UUID, error)
 
 type operationState struct {
 	Ready      string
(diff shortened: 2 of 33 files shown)


xuzhang3:
@rdalbuquerque Since this is an AzureRM service connection secret rotate issue, suggest use API(https://learn.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpointproxy/execute-service-endpoint-request?tabs=HTTP) to do the validation for service connection.  Currently all service connections not do the validation by default which means all secrets can be dummy values. Add the retry logic in VG only resolve part of the issue as other resource may also face the same issue.

Web UI Verify:
image

rdalbuquerque:
@xuzhang3  Ok, so I played around a few scenarios to make this validation possible and more generic:

```go
func validateServiceEndpointConnection(clients *client.AggregatedClient, endpoint *serviceendpoint.ServiceEndpoint) error {
	reqArgs := serviceendpoint.ExecuteServiceEndpointRequestArgs{
		ServiceEndpointRequest: &serviceendpoint.ServiceEndpointRequest{
			DataSourceDetails: &serviceendpoint.DataSourceDetails{
				DataSourceName: converter.String("TestConnection"),
			},
			ResultTransformationDetails: &serviceendpoint.ResultTransformationDetails{},
		},
		Project:    converter.String((*endpoint.ServiceEndpointProjectReferences)[0].ProjectReference.Id.String()),
		EndpointId: converter.String(endpoint.Id.String()),
	}

	maxRetries := 15  // should probably be an input
	retryInverval := 2  // should probably be an input
	var reqResult *serviceendpoint.ServiceEndpointRequestResult
	var err error
	log.Printf(fmt.Sprintf(":: serviceendpoint/commons.go :: %s :: Initiating validation", *endpoint.Name))
	for i := 0; i < maxRetries; i++ {
		if reqResult, err = clients.ServiceEndpointClient.ExecuteServiceEndpointRequest(clients.Ctx, reqArgs); err != nil {
			log.Printf(fmt.Sprintf(":: serviceendpoint/commons.go :: %s :: error during endpoint validation request", *endpoint.Name))
			return err
		} else if strings.EqualFold(*reqResult.StatusCode, "ok") {
			log.Printf(fmt.Sprintf(":: serviceendpoint/commons.go :: %s :: successfully validated connection", *endpoint.Name))
			return nil
		}
		log.Printf(fmt.Sprintf(":: serviceendpoint/commons.go :: %s :: validation failed, retrying...", *endpoint.Name))
		time.Sleep(time.Duration(retryInverval) * time.Second)
	}

	return fmt.Errorf("Error validating connection: (type: %s, name: %s, code: %s, message: %s)", *endpoint.Type, *endpoint.Name, *reqResult.StatusCode, *reqResult.ErrorMessage)
}
```
1. Run the function above **everytime** **any** serviceendpoint is created or updated This is problematic because:
    - To run this proxy request I need the EndpointID which I only get after endpoint is successfully created and that can lead to the endpoint having to be deleted because validation didn't pass and I don't know if that's a good practice in plugin development.
    - I don't know if every endpoint has this validation functionality implemented, through the UI it looks like it's not and it even depends on the authorization method.

2. Run this validation conditioned to the endpoint type (hardcoded). there would somehow be a parameter to tell the provider whether specified endpoint type has to run the validation or not.
problem:
    - Even if I know the service endpoint type support validation (azurerm case) I might be breaking use cases where people actually want for some reason create an invalid endpoint (maybe access is given afterwards).

3. Create a new argument on resource schema telling the provider whether or not that endpoint should be validated after creation/update. I don't see any issue here, I don't know if that's the best way to handle this
4. Add this function to helper package where it could be called from other packages (like variable group resource). I would just need to change the function inputs so it won't require the full endpoint object

Hopefully this can be solved. Thanks in advance!

xuzhang3:
@rdalbuquerque 
1. Run the function above everytime any serviceendpoint is created or updated This is problematic because:
   - Validation will not delete the service endpoint, it just verify the connection.
   -  Not all the endpoint implement the validation function, only some have.

2. This could be flag, defaults to `false`. Only do the validation when it is `true`.  
3. Validation is a feature of some endpoints. In this scenario the AD secret changed, the real affected resource is `azuredevops_serviceendpoint_azurerm`, `azuredevops_serviceendpoint_azurerm` should ensure it works not `azuredevops_variable_group` 
4. Add this to the helper package/common package is helpful, any endpoint that supports this feature can use it

rdalbuquerque:
@xuzhang3 
OK so here are some toughts on the new changes I pushed:
I had a hard time isolating the changes to just `resource_serviceendpoint_azurerm.go` since only flatten and expand functions are delegated to specific endpoint files and each endpoint was expeted to return a specific type of function that didn't have an input to fit validate argument.

So for this first attempt I did the following:
1. changed the expand/flattenFunc types to handle a new struct `serviceEndpointWithValidation` instead of `serviceendpoint.ServiceEndpoint` so I could return if the azurerm endpoint should be validated or not on create/update operations
```go
type serviceEndpointWithValidation struct {
	endpoint *serviceendpoint.ServiceEndpoint
	validate bool
}
type flatFunc func(d *schema.ResourceData, serviceEndpoint *serviceEndpointWithValidation, projectID *uuid.UUID)
type expandFunc func(d *schema.ResourceData) (*serviceEndpointWithValidation, *uuid.UUID, error)
```
2. Because of that I had to adapt all the other endpoints to handle `serviceEndpointWithValidation` type (and all their tests)
3. Added a conditional on `genServiceEndpointCreateFunc` and `genServiceEndpointUpdateFunc` to handle validation in case validate argument is set to `true`
    - on `genServiceEndpointCreateFunc` I can only run validation after endpoint is created so I can have the endpoint id, because of that, if validation fails I then delete the endpoint and return an error
    - on `genServiceEndpointUpdateFunc` I'm able to validate the new configs before applying changes, so if validation fails I return an error and service endpoint object remains unchanged
4. added the unit and acceptance tests to cover the new `validate` argument

One issue I couldn't figure out the best way to handle is that even if the update fails and, on the update, I'm changing the `validate` argument, the new `validate` value is still saved to state because on `genServiceEndpointReadFunc` doesn't have a way to know the actual `validate` value since it's only on state and not an object property in Azure Devops, so I get the value from config:
```go
validate, ok := d.Get("validate").(bool)
if !ok {
  validate = false
}
flatFunc(d, &serviceEndpointWithValidation{endpoint: serviceEndpoint, validate: validate}, &projectID)
```

Looking forward to you feedback :)

xuzhang3:
@rdalbuquerque  currently all the service connections share the same create/update/read/delete function which is a limitation to all the service connection resources,  some service connections have specifical configurations which is only available to itself. We are considering refactoring the service connection resources, every resource has its own create/update/read/delete functions but share the common functions. Each resource can customize its own configuration without affecting other resources.

rdalbuquerque:
@xuzhang3 I see, so we wait the refactor before moving forward with this PR?

If there is anything I can do to help refactoring or any other way we can implement this validation on azurerm service endpoint that would not need the refactor,  please let me know!

xuzhang3:
@rdalbuquerque yes, wait before refactoring. More work is required if this PR is merged.

xuzhang3:
@rdalbuquerque  #863 will decouple the generic functions, you can add the validation to `azurerm` after that PR merged

xuzhang3:
@rdalbuquerque  #863 has been merged, you can add the validation to `azurerm` resoruce without affecting other resources

PR #881, opened 2023-09-08
Setup conflict checks for variable secret/non-secret attributes in VG

AlexPykavy:
To throw an error if the `value` attribute is used when secret was expected, and vice versa.

## All Submissions:

* [ ] My code follows the code style of this project.
* [ ] Have you checked to ensure there aren't other open PRs for the same update/change?

Issue Number:

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

### DIFF
--- a/azuredevops/internal/service/taskagent/resource_variable_group.go
+++ b/azuredevops/internal/service/taskagent/resource_variable_group.go
@@ -108,20 +108,20 @@ func ResourceVariableGroup() *schema.Resource {
 							Type:          schema.TypeString,
 							Optional:      true,
 							Default:       "",
-							ConflictsWith: []string{vgKeyVault},
+							ConflictsWith: []string{vgKeyVault, vgIsSecret, secretVgValue},
 						},
 						secretVgValue: {
 							Type:          schema.TypeString,
 							Optional:      true,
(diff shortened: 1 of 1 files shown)


azuredevops/internal/service/taskagent/resource_variable_group.go
```diff
@@ -108,20 +108,20 @@ func ResourceVariableGroup() *schema.Resource {
 							Type:          schema.TypeString,
 							Optional:      true,
 							Default:       "",
-							ConflictsWith: []string{vgKeyVault},
+							ConflictsWith: []string{vgKeyVault, vgIsSecret, secretVgValue},
```
xuzhang3, line 111: `value, secret_value, is_secret` all have default values, the constraint will always  be valid, internal check will always failed

AlexPykavy: You are absolutely right. Moreover, referencing objects in the Set is [not supported](https://github.com/hashicorp/terraform-plugin-sdk/issues/71) in `ConflictsWith` at the moment. I have another idea how to achieve this but it will be more dangerous and deserves a new PR.

PR #887, opened 2023-09-15
Don't use `is_secret` attribute for secrets in variable groups

AlexPykavy:
Because specifying the `secret_value` attribute is enough to designate the variable as secret.

## All Submissions:

Issue Number:

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/azuredevops/internal/acceptancetests/testutils/hcl.go
+++ b/azuredevops/internal/acceptancetests/testutils/hcl.go
@@ -646,7 +646,6 @@ resource "azuredevops_variable_group" "vg" {
 	variable {
 		name      = "key1"
 		secret_value  = "value1"
-		is_secret = true
 	}
 
 	variable {
--- a/azuredevops/internal/service/taskagent/resource_variable_group.go
+++ b/azuredevops/internal/service/taskagent/resource_variable_group.go
@@ -118,6 +118,7 @@ func ResourceVariableGroup() *schema.Resource {
 							ConflictsWith: []string{vgKeyVault},
 						},
 						vgIsSecret: {
+							Deprecated:    fmt.Sprintf("Don't use %s anymore as specifying %s is enough", vgIsSecret, secretVgValue),
 							Type:          schema.TypeBool,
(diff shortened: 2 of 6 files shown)


xuzhang3:
@AlexPykavy Thank you for opening this PR. Is there any specific reason why to remove the `is_secret` property?

AlexPykavy:
Because having the `value`, `is_secret` and `secret_value` triplet is redundant and confusing. If you want to define the secret, you need to first set two attributes: `is_secret` and `secret_value`.

For me, the more logical approach is to use binary logic: either `value`, or `secret_value`.

xuzhang3:
You are right, `value` or `secret_value` is enough here.  We originally introduced `is_secret` into the ADO provider for mapping to API parameters.

HerrSubset:
This would be a great change. I just wasted waaay too much time trying to figure out why some secret variable was empty when using `value = xxx` in combination with `is_secret = true`. 

Sure, it's documented and examples are available, but when you have worked with this provider only a couple of times, you might remember the _is_secret_ property, but not necessarily that you also need to switch to _secret_value_. And then it gets confusing quickly...

Instead of removing the _is_secret_ property, you could also remove _secret_value_, then the change doesn't have to be breaking immediately. For regular variables, users keep using _value_. For a secret variable, you can use either _value_ or _secret_value_ in combination with _is_secret_. But if you use _secret_value_, you get a warning, saying it's deprecated and you should use _value_ instead?

Then after some time, you can remove _secret_value_ in a breaking change. But at least there was a transition period.


Or another option: display an error when `is_secret = true` is defined without a _secret_value_.

AlexPykavy:
Hi @HerrSubset , thank you for the upvoting. I've updated this PR to make it a little clearer. To be honest, even I had a hard time understanding some things since it was opened a long time ago 😆

I'll keep this one as it seems as a proper solution to the problem. At the same time, I agree that adding simple validation will help users of this provider right now, so I'll open a new one.

Issue #877, opened 2023-09-06
Create data source for azuredevops_serviceendpoint_kubernetes


muresan:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Create a data source for azuredevops_serviceendpoint_kubernetes, this would help discovering the resource id by the project_id and endpoint name to discover endpoint_id or by endpoint_id and project name to discover service_endpoint_name.

### New or Affected Resource(s)

* data.azuredevops_serviceendpoint_kubernetes

### Potential Terraform Configuration

```hcl
data "azuredevops_serviceendpoint_kubernetes" "this" {
  project_id            = "00000000-0000-0000-0000-000000000000"
  service_endpoint_name = "example-service-endpoint-kubernetes"
}

output "service_endpoint_id" {
  value = data.azuredevops_serviceendpoint_kubernetes.this.id
}

```

muresan:
Resolved by https://github.com/microsoft/terraform-provider-azuredevops/pull/1400

PR #926, opened 2023-11-10
Added support for data_serviceendpoint_kubernetes

muresan:
## All Submissions:

I did not see tests added to the other data sources, so I did not add one for this one.

The linked issue has info about this PR

Issue Number: https://github.com/microsoft/terraform-provider-azuredevops/issues/877

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/serviceendpoint/data_serviceendpoint_kubernetes.go
+++ b/azuredevops/internal/service/serviceendpoint/data_serviceendpoint_kubernetes.go
@@ -1,0 +1,24 @@
+package serviceendpoint
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataResourceServiceEndpointKubernetes() *schema.Resource {
+	return dataSourceGenBaseServiceEndpointResource(dataSourceServiceEndpointKubernetesRead)
+}
+
+func dataSourceServiceEndpointKubernetesRead(d *schema.ResourceData, m interface{}) error {
+	serviceEndpoint, projectID, err := dataSourceGetBaseServiceEndpoint(d, m)
+	if err != nil {
+		return err
+	}
+	if serviceEndpoint != nil {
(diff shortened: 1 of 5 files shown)


muresan:

xuzhang3 (changes requested):
@@muresan Thank you for your contribute. Can you help add AccTest to cover the new data source. And add separate schema(data source properties are simply `compute=true`)  and flatten function for data source not simply reuse the resource functions.

muresan:
> @@muresan Thank you for your contribute. Can you help add AccTest to cover the new data source. And add separate schema(data source properties are simply `compute=true`) and flatten function for data source not simply reuse the resource functions.

Thank you for your comment. I added the kubernetes service endpoint in the same way as the sonar cloud service endpoint was added. Can you give me an example of another data source that I should follow ? All service endpoints data sources are added in the same way.
As for tests, can you point me to an example ? All the other serivice endpoint data souces had no tests so my assumption was that they are not desired. Can you point me to an example ?

Thanks.

xuzhang3:
@muresan  Data source only need the required parameter to get the service connection, other properties are not required. In the scenario, you are using the common schema for the k8s service connection. This is OK as the common schema cover the minimal properties requirement. You can add the same properties as `azuredevops_serviceendpoint_kubernetes` and all these properties are `computed=true`
Example:
```go
r.Schema["service_account"] = &schema.Schema{
		Type:     schema.TypeList,
		MaxItems: 1,
		Computed: true,
		Elem: map[string]*schema.Schema{
			"ca_cert": {
				Type:     schema.TypeString,
				Computed: true,
			},
			"token": {
				Type:     schema.TypeString,
				Computed: true,
			},
		},
	}
```
You can ref:  https://github.com/microsoft/terraform-provider-azuredevops/blob/main/azuredevops/internal/service/build/data_build_definition.go

muresan:
This has been resolved by https://github.com/microsoft/terraform-provider-azuredevops/pull/1400

PR #939, opened 2023-12-13
Improve read operation performance of `azuredevops_users` data source

rdalbuquerque:
## All Submissions:

While using `azuredevops_users` data source to fetch all users, or even `svc` subject type users, the reading operation can take too long depending on the number of users on the organization. 

So to improve performance, it was added a `features` block with `concurrent_workers` argument so the user can set concurrent processing for the user data source, greatly decreasing the time it takes to finish read operation.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Without concurrent workers:

image

With `concurrent_workers` set to 10:

image

## Other information

`azuredevops_users` data source with features block:
```hcl
data "azuredevops_users" "users" {
  features {
    concurrent_workers = 10
  }
}
```


xuzhang3 (approved):

drdamour:
pretty concerned this can make things even worse..the underlying problem with azuredevops_users is it pulls every user in cases you only want 1

rdalbuquerque:
@drdamour on my use cases it either drastically improved performance (like the print on this PR) or stayed the same, and default is a single go routine.

Feel free to contact me with any use case that this might be negatively impacting and we can discuss possible solutions.

xuzhang3:
@rdalbuquerque  @drdamour This can be controlled by adding a custom attribute to control the number of users that can be pulled.  Similar like the `azuredevops_variable_group`, a custom property `search_depth` was used to control how many `secrets` can be searched.

Issue #976, opened 2024-02-21
Changing requester_can_approve on azuredevops_check_approval causes an error


josh-barker:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform v1.7.0
on darwin_amd64
+ provider registry.terraform.io/microsoft/azuredevops v0.10.0

+ provider registry.terraform.io/microsoft/azuredevops v0.11.0

+ provider registry.terraform.io/microsoft/azuredevops v0.9.1

### Affected Resource(s)

* `azuredevops_check_approval`

### Terraform Configuration Files

```hcl

resource "azuredevops_project" "example" {
  name = "Example Project"
}

resource "azuredevops_environment" "example" {
  project_id = azuredevops_project.example.id
  name       = "Example Environment"
}

resource "azuredevops_group" "example" {
  display_name = "some-azdo-group"
}

resource "azuredevops_check_approval" "example" {
  project_id           = azuredevops_project.example.id
  target_resource_id   = azuredevops_environment.example.id
  target_resource_type = "environment"

  requester_can_approve = false
  approvers = [
    azuredevops_group.example.origin_id,
  ]
}

```

### Actual Behavior

```shell
Terraform will perform the following actions:

  # azuredevops_check_approval.example will be updated in-place
  ~ resource "azuredevops_check_approval" "example" {
        id                         = "736"
      ~ requester_can_approve      = false -> true
        # (6 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
azuredevops_check_approval.example: Modifying... [id=736]
╷
│ Error: Specified argument was out of the range of valid values.
│ Parameter name: Version
│ 
│   with azuredevops_check_approval.example,
│   on main.tf line 314, in resource "azuredevops_check_approval" "example":
│  314: resource "azuredevops_check_approval" "example" {
│ 
╵
```

### Steps to Reproduce

1. `terraform apply`
2. Set `requester_can_approve` to `true`
3. `terraform apply`

I've tried a few versions of the provider, and all have the same error.
[...]

PR #977, opened 2024-02-22
fix: add missing version property on check resources

josh-barker:
## All Submissions:

From my testing, the rest API in AzureDevOps has introduced a new required fields `version`, although the API documentation doesn't mention it.

https://learn.microsoft.com/en-us/rest/api/azure/devops/approvalsandchecks/check-configurations/update?view=azure-devops-rest-7.1&tabs=HTTP

I've added a new computed property to read and send it from/to AZDO.

Issue Number: #976 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

You can see the existing bug by running tests on `^TestAccCheck\\w+_update$` on the current version of the provider.

### DIFF
--- a/azuredevops/internal/service/approvalsandchecks/common_check.go
+++ b/azuredevops/internal/service/approvalsandchecks/common_check.go
@@ -86,6 +86,10 @@ func genBaseCheckResource(f flatFunc, e expandFunc) *schema.Resource {
 				ForceNew:     true,
 				ValidateFunc: validation.StringInSlice(targetResourceTypes, false),
 			},
+			"version": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
 		},
 	}
 }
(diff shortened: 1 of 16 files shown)


azuredevops/internal/service/approvalsandchecks/common_check.go
```diff
@@ -101,6 +105,7 @@ func doBaseExpansion(d *schema.ResourceData, checkType *pipelineschecksextras.Ch
 			Id:   converter.String(d.Get("target_resource_id").(string)),
 			Type: converter.String(d.Get("target_resource_type").(string)),
 		},
+		Version: converter.Int(d.Get("version").(int)),
```
xuzhang3, line 108: Since `version` is computed we don't need to send it to the service

xuzhang3 (changes requested):

azuredevops/internal/service/approvalsandchecks/common_check.go
josh-barker: @xuzhang3 When I remove that line of code, the API has the same error.

image


On a create, the API doesn't require a `version` property in the payload, but it returns a `version` property.
On an update, the API requires a `version` property, which is a whole integer which increments each time the resources is updated.

xuzhang3 (approved):

Issue #979, opened 2024-02-23
azuredevops_team_administrators does not set correctly permissions


lukaskolafa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

OpenTofu v1.6.1
provider registry.opentofu.org/microsoft/azuredevops v0.11.0

### Affected Resource(s)

* `azuredevops_team_administrators`
* `azuredevops_team` (administrators property)

### Terraform Configuration Files

```hcl
resource "azuredevops_team" "additional_team" {
  project_id  = azuredevops_project.this.id
  name        = "something"
  description = "something"
}

resource "azuredevops_team_administrators" "additional_team" {
  project_id     = azuredevops_project.this.id
  team_id        = azuredevops_team.additional_team.id
  mode           = "add"
  administrators = [....]
}
```

### Debug Output

not required

### Panic Output

not required

### Expected Behavior

The listed administrators in hcl should become administrators of the created DevOps Team. The permission flag should be `allow-bit: 31` in the identity namespace `5a27515b-ccd7-42c9-84f1-54c998f03866`

### Actual Behavior

The listed administrators in hcl are created in DevOps Team and even the UI displays them as administrators. But behind the scenes, their permission flag is `allow-bit: 8`, which is sufficient only for managing members of the Team, but does not work for managing administrators.

### Steps to Reproduce

1. `tofu apply`
2. check the Team in DevOps -> administrators are displayed
3. administrator tries to add new member to the team -> works
4. administrator tries to add new administrator to the team -> does not work
5. remove and re-add the administrator in the UI (using a Global Administrator)
6. repeat step 4 -> now it works

### Important Factoids

N/A

### References

* https://github.com/microsoft/terraform-provider-azuredevops/blob/f780a17e8ac7fc83fb829aa8153a569cb9d5a972/azuredevops/internal/service/core/resource_team.go#L566

I assume the bug is in the file resource_team.go (line 566). The list of permissions is set to `ManageMebership` only. The correct list should be `["Read", "Write", "Delete", "ManageMembership", "CreateScope"]`

[...]

PR #981, opened 2024-02-27
`azuredevops_team` - Fixing the team admin flags

lukaskolafa:
## All Submissions:

Before: Team administrator was created with flags ManageMembership
After: Team administrator is created with flags Read, Write, Delete, ManageMembership, CreateScope

The behavior before is not consistent with the UI. When you create a team administrator via UI, it is consistent with the behavior after.

Issue Number: #979 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/core/resource_team.go
+++ b/azuredevops/internal/service/core/resource_team.go
@@ -494,11 +494,10 @@ func readTeamAdministrators(d *schema.ResourceData, clients *client.AggregatedCl
 
 	adminDescriptorList := []string{}
 	if acl != nil && acl.AcesDictionary != nil {
-		// For backwards compatibility, we check only the ManageMembership bit, but a proper admin actually requires all bits filled (ace.Allow=31). When creating a new admin,
-		// all bits are set, so this is not a problem.
-		bit := *(*actionDefinitions)["ManageMembership"].Bit
+		bit := *(*actionDefinitions)["Read"].Bit | *(*actionDefinitions)["Write"].Bit | *(*actionDefinitions)["Delete"].Bit | *(*actionDefinitions)["ManageMembership"].Bit | *(*actionDefinitions)["CreateScope"].Bit
 		for _, ace := range *acl.AcesDictionary {
-			if *ace.Allow&bit > 0 {
+
+			if *ace.Allow&bit == bit {
 				adminDescriptorList = append(adminDescriptorList, *ace.Descriptor)
 			}
 		}
--- a/azuredevops/internal/service/core/resource_team.go
+++ b/azuredevops/internal/service/core/resource_team.go
@@ -494,6 +494,8 @@ func readTeamAdministrators(d *schema.ResourceData, clients *client.AggregatedCl
 
 	adminDescriptorList := []string{}
 	if acl != nil && acl.AcesDictionary != nil {
+		// For backwards compatibility, we check only the ManageMembership bit, but a proper admin actually requires all bits filled (ace.Allow=31). When creating a new admin,
+		// all bits are set, so this is not a problem.
 		bit := *(*actionDefinitions)["ManageMembership"].Bit
 		for _, ace := range *acl.AcesDictionary {
 			if *ace.Allow&bit > 0 {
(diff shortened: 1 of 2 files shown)


lukaskolafa:

xuzhang3:
@lukaskolafa  lint check failed
```log
==> Checking that code complies with gofmt requirements...
gofmt needs running on the following files:
./azuredevops/internal/service/core/resource_team_test.go
You can use the command: `make fmt` to reformat code.
make: *** [GNUmakefile:36: fmtcheck] Error 1

```

lukaskolafa:
thanks for the hint @xuzhang3 , I was a bit unsure with that command, because it changed 350 files in my dev environment. Anyway, I pushed the changes and it looks like it was only a whitespace change, so it should look better now.

One thing we should discuss is the backward compatibility with the old behavior. The current implementation recognizes admin by the old flag and new admins are created with the full list of flags. I believe this will work in praxis, but it is not a 100% correct implementation. What do you suggest as a resolution for this problem?

xuzhang3:
@lukaskolafa  The difference should be caused by the line-endings characters of different platforms.(CRLF, LF).   
Not tested. Will this fix trigger an update of existing resources since permissions have changed? If so the update should help set the permission correctly.

lukaskolafa:
@xuzhang3  I updated the code to recognize correctly admins only if all 5 flags are set. In case the resource is tainted, it will properly do an update. Unfortunately, Terraform does not recognize the difference on its own without tainting, which I believe is a standard Terraform behavior.

azuredevops/internal/service/core/resource_team.go
```diff
@@ -563,7 +564,11 @@ func setTeamAdministratorsPermissions(d *schema.ResourceData, clients *client.Ag
 				PrincipalPermission: securityhelper.PrincipalPermission{
 					SubjectDescriptor: item.(string),
 					Permissions: map[securityhelper.ActionName]securityhelper.PermissionType{
+						"Read":             permission,
+						"Write":            permission,
+						"Delete":           permission,
 						"ManageMembership": permission,
+						"CreateScope":      permission,
```
xuzhang3, line 570: Checked  the default team admin and it have 6 permissions:
```
0 = Read -> 
1 = Write -> 
2 = Delete -> 
3 = ManageMembership -> 
4 = CreateScope -> 
5 = RestoreScope -> 
```

xuzhang3 (changes requested):

azuredevops/internal/service/core/resource_team.go
lukaskolafa: how exactly did you test this? I can confirm, that the `RestoreScope` bit exists, this command

`az devops security permission namespace list --organization "https://dev.azure.com/...."` gives me this result:

```
...
[
      {
        "bit": 1,
        "displayName": "View identity information",
        "name": "Read",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      },
      {
        "bit": 2,
        "displayName": "Edit identity information",
        "name": "Write",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      },
      {
        "bit": 4,
        "displayName": "Delete identity information",
        "name": "Delete",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      },
      {
        "bit": 8,
        "displayName": "Manage group membership",
        "name": "ManageMembership",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      },
      {
        "bit": 16,
        "displayName": "Create identity scopes",
        "name": "CreateScope",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      },
      {
        "bit": 32,
        "displayName": "Restore identity scopes",
        "name": "RestoreScope",
        "namespaceId": "5a27515b-ccd7-42c9-84f1-54c998f03866"
      }
    ]
...
```

but, when I create a new Project and check the permissions using this command (PowerShell):

`$namespacePermissions = Invoke-RestMethod -Uri "$organizationUri/_apis/accesscontrollists/$($namespace.namespaceId)?api-version=7.1-preview.1" -Method GET -Headers $authenticationHeader`

Then I retrieve the value `allow = 31` for the default Team administrator, which seems to be the value stored in AZDO DB.

xuzhang3: You are right, the admin allow value is 31.

xuzhang3: Lint check failed:
```log
azuredevops/internal/service/core/resource_team.go:498:44: unnecessary leading newline (whitespace)
```

lukaskolafa: sorry, I fixed the lines

xuzhang3 (approved):

PR #1075, opened 2024-06-14
`azuredevops_variable_group` - Add validation that variable can have either only `value` attribute or both `is_secret` and `secret_value` attributes

AlexPykavy:
A temporary fix till https://github.com/microsoft/terraform-provider-azuredevops/pull/887 or any better solution is merged.

## All Submissions:

Issue Number:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/service/taskagent/resource_variable_group.go
+++ b/azuredevops/internal/service/taskagent/resource_variable_group.go
@@ -375,6 +375,21 @@ func expandVariableGroupParameters(clients *client.AggregatedClient, d *schema.R
 	description := converter.String(d.Get(vgDescription).(string))
 	variables := d.Get(vgVariable).(*schema.Set).List()
 
+	// needed to detect if the secret_value attribute is set in the config
+	// see https://github.com/hashicorp/terraform-plugin-sdk/issues/741
+	for it := d.GetRawConfig().AsValueMap()[vgVariable].ElementIterator(); it.Next(); {
+		_, ctyVariable := it.Element()
+		ctyVariableAsMap := ctyVariable.AsValueMap()
+		name := ctyVariableAsMap[vgName].AsString()
+		valueSet := !ctyVariableAsMap[vgValue].IsNull()
+		secretValueSet := !ctyVariableAsMap[secretVgValue].IsNull()
+		isSecretSet := !ctyVariableAsMap[vgIsSecret].IsNull()
+
+		if valueSet && (secretValueSet || isSecretSet) || secretValueSet != isSecretSet {
+			return nil, nil, fmt.Errorf("`%s` variable can have either only `value` attribute or both `is_secret` and `secret_value` attributes", name)
+		}
+	}
+
 	variableMap := make(map[string]interface{})
 
 	for _, variable := range variables {
--- a/website/docs/r/variable_group.html.markdown
+++ b/website/docs/r/variable_group.html.markdown
@@ -101,6 +101,8 @@ The following arguments are supported:
 
 A `variable` block supports the following:
 
+!> **Warning** variable can have either only `value` attribute or both `is_secret` and `secret_value` attributes
+
 - `name` - (Required) The key value used for the variable. Must be unique within the Variable Group.
 - `value` - (Optional) The value of the variable. If omitted, it will default to empty string.
 - `secret_value` - (Optional) The secret value of the variable. If omitted, it will default to empty string. Used when `is_secret` set to `true`.


azuredevops/internal/service/taskagent/resource_variable_group.go
```diff
@@ -375,6 +375,21 @@ func expandVariableGroupParameters(clients *client.AggregatedClient, d *schema.R
...
+	for it := d.GetRawConfig().AsValueMap()[vgVariable].ElementIterator(); it.Next(); {
+		_, ctyVariable := it.Element()
+		ctyVariableAsMap := ctyVariable.AsValueMap()
+		name := ctyVariableAsMap[vgName].AsString()
+		valueSet := !ctyVariableAsMap[vgValue].IsNull()
+		secretValueSet := !ctyVariableAsMap[secretVgValue].IsNull()
+		isSecretSet := !ctyVariableAsMap[vgIsSecret].IsNull()
+
+		if valueSet && (secretValueSet || isSecretSet) || secretValueSet != isSecretSet {
+			return nil, nil, fmt.Errorf("`%s` variable can have either only `value` attribute or both `is_secret` and `secret_value` attributes", name)
```
HerrSubset, line 389: I think it might be clearer for the user if you split this up in 2 error cases/messages:
1. you can't set both `value` and `secret_value`
2. If `secret_value` is set, you must set `is_secret = true`

I think a combination of these two cases/messages will guide the user to the correct configuration for their variable. And the messages can be a bit clearer than now, trying to explain the relationship between the 3 in one sentence.

EDIT: I was a bit too fast there, this doesn't cover setting `is_secret = true` in combination with setting `value`...

xuzhang3 (approved):

emibcn:
@xuzhang3 @AlexPykavy @HerrSubset 

Is there any way to **conditionally** set `value` or `secret_value`?

I have a Terraform module to define AzDo VariableGroups, and in it I use a `for_each` reading the values from an object:
```
value        = coalesce(lookup(variable.value, "is_secret", false), false) ? null : variable.value.value
secret_value = coalesce(lookup(variable.value, "is_secret", false), false) ? variable.value.value : null
is_secret    = coalesce(lookup(variable.value, "is_secret", false), false)
```

And with the changes in this PR now it always throws the error:
```
│ Error: Expanding variable group resource data: `SOME_VARIBLE_NAME` variable can have either only `value` attribute or both `is_secret` and `secret_value` attributes
```

Despite `value` or `secret_value` are set to `null` which, in Terraform, is normally used as `non-set attribute`. I have also tested setting the non-set attribute to empty string `""` instead, but with the same error message being thrown.

So I guess that instead of only looking for the attribute being set, you should also look for it's value if set and, if it's not accepted, then throw the error:
- If `is_secret` is `false`, it should act as if not set
- If `secret_value` is `null`, it should act as if not set (beware that documentation states that not setting this will act as an empty string!!)
- If `value` is `null`, it should act as if not set (beware that documentation states that not setting this will act as an empty string!!)

image
https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/variable_group#argument-reference

**Note:** The last version where it worked was **microsoft/azuredevops v1.1.1**

AlexPykavy:
@emibcn , I've found a workaround for your issue:
```terraform
variable "variables" {
  type = map(object({
    value        = string
    secret_value = string
    is_secret    = bool
  }))
  default = {
    plain_var = {
      value        = "plain_val"
      secret_value = null
      is_secret    = false
    }
    secret_var = {
      value        = null
      secret_value = "secret_val"
      is_secret    = true
    }
  }
}

terraform {
  required_providers {
    azuredevops = {
      source = "microsoft/azuredevops"
    }
  }
}

provider "azuredevops" {
  org_service_url       = ""
  personal_access_token = ""
}

data "azuredevops_project" "main" {
  name = "Playground"
}

resource "azuredevops_variable_group" "main" {
  project_id   = data.azuredevops_project.main.id
  name         = "Example Variable Group"
  description  = "Example Variable Group Description"
  allow_access = true

  dynamic "variable" {
    for_each = { for k, v in var.variables : k => v if !v.is_secret }
    content {
      name  = variable.key
      value = variable.value.value
    }
  }

  dynamic "variable" {
    for_each = { for k, v in var.variables : k => v if v.is_secret }
    content {
      name         = variable.key
      secret_value = variable.value.secret_value
      is_secret    = true
    }
  }
}
```

emibcn:
Hi @AlexPykavy ,
Thanks for the tip. Yes, that's a workaround I will definitely use (or something similar 😉 ). But that's a workaround, it should be fixed...

PR #1351, opened 2025-04-11
[Bug Fix] - `azuredevops_serviceendpoint_azurerm` - Fix bug with serviceendpoint_azurerm when shared projectid gets mutil…

leasley199:
bug with serviceendpoint_azurerm  when shared projectID gets clobbered/mutilated. 

## All Submissions:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] I have updated the documentation accordingly.
* [X] I have added tests to cover my changes.
* [X] All new and existing tests passed.
* [X] My code follows the code style of this project.
* [X] I ran lint checks locally prior to submission.
* [X] Have you checked to ensure there aren't other open PRs for the same update/change?

Currently there is a bug with the serviceendpoint_azurerm resource when the service connection gets shared to another project the projectid gets clobbered. this is because the value for the projectID is hardcoded to [0] and the object that this id is read from the order or IDs can change and thus break the resource. 

```
[{"projectReference":{"id":"0000-id00001","name":"Project01"},"name":"AZDO-Project01","description":"Managed by Terraform"},{"projectReference":{"id":"0000-id00002","name":"Project02"},"name":"AZDO-Project02","description":"Managed by Terraform"}]
```

if the shared projectID is alphabetically/numerically greater than the initial projectID this will cause a mismatch and thus the resource is broken. The fix takes the field, matches the specified projected of the resource and continues if successful/else fail if not match (obviously ha ha).

 This is probably an issue in all serviceendpoints that have the projectID hardcoded as [0] but in the effort of do no harm I am only targeting this resource. 
The changes 'should' be able to lift and shift however if true. 

Issue Number: NA

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

current error tries to delete and recreate resource since projectID changes. This is undesired since it breaks the shared service connection resource and state becomes undesirable/falsified.

Test locally, but more eyes and insight never hurt many thanks :)

### DIFF
--- a/azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
+++ b/azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/microsoft/azure-devops-go-api/azuredevops/v7/serviceendpoint"
@@ -252,12 +251,6 @@ func resourceServiceEndpointAzureRMRead(d *schema.ResourceData, m interface{}) e
 		return err
 	}
 
-	projectID := d.Get("project_id").(string)
-    _, err = validateProjectID(d, projectID, m)
-    if err != nil {
-        return fmt.Errorf("validation failed: %v", err)
-    }
-	
 	serviceEndpoint, err := clients.ServiceEndpointClient.GetServiceEndpointDetails(clients.Ctx, *getArgs)
 	if isServiceEndpointDeleted(d, err, serviceEndpoint, getArgs) {
 		return nil
@@ -619,34 +612,4 @@ func shouldValidate(features map[string]interface{}) bool {
 		return false
 	}
 	return validate
-}
-
-
-func validateProjectID(d *schema.ResourceData, projectID string, m interface{}) (*uuid.UUID, error) {
-	clients := m.(*client.AggregatedClient)
-	getArgs, err := serviceEndpointGetArgs(d)
-	if err != nil {
-		return nil, err
-	}
-	var matchedProjectID *uuid.UUID
-	serviceEndpoint, err := clients.ServiceEndpointClient.GetServiceEndpointDetails(clients.Ctx, *getArgs)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, ref := range *serviceEndpoint.ServiceEndpointProjectReferences {
-		if ref.ProjectReference != nil && ref.ProjectReference.Id != nil &&
-			ref.ProjectReference.Id.String() == projectID {
-			matchedProjectID = ref.ProjectReference.Id
-			break
-		}
-	}
-
-	if matchedProjectID == nil {
-		return nil, fmt.Errorf("project_id '%s' not found in service endpoint project references", projectID)
-	}
-
-	// Set the project_id in the resource data for use in other functions
-	d.Set("project_id", matchedProjectID.String())
-	return matchedProjectID, nil
 }
--- a/azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
+++ b/azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
@@ -172,6 +172,14 @@ func ResourceServiceEndpointAzureRM() *schema.Resource {
 				},
 			},
 		},
+		"shared_project_ids": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			Description: "A list of additional project IDs where the service endpoint should be shared.",
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
 	})
 
 	r.SchemaVersion = 2
@@ -198,6 +206,12 @@ func resourceServiceEndpointAzureRMCreate(d *schema.ResourceData, m interface{})
         return fmt.Errorf(errMsgTfConfigRead, err)
     }
 
+	projectID := d.Get("project_id").(string)
+    _, err = validateProjectID(d, projectID, m)
+    if err != nil {
+        return fmt.Errorf("validation failed: %v", err)
+    }
+
     resp, err := createServiceEndpoint(d, clients, serviceEndpoint)
     if err != nil {
         return err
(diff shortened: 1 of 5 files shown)


xuzhang3:
@leasley199 Currently the `project_id` is being checked during the read(https://github.com/microsoft/terraform-provider-azuredevops/blob/main/azuredevops/internal/service/serviceendpoint/commons.go#L311).  You should add this check to the `read` operation instead of the `create` operation. Also the `validateServiceEndpoint` should not removed, it is used to validate the service connection after setting the configuration and is the same as `Verify and save`. 
image

leasley199:
weird I can see the validateServiceEndpoint on my local branch untouched. noted on the read/create operation.

xuzhang3:
> weird I can see the validateServiceEndpoint on my local branch untouched. noted on the read/create operation.

Wrong PR? `validateServiceEndpoint` has been removed in this PR.
image

leasley199:
alright new commit(s) should address the prior comment. I also added a parameter to the resource that cause the initial bug in the first place. When service connection is shared project Id was getting clobbered, now projectID will validate as well if the serv connection is shared with other projects this can be managed via tf now. If not included serv connection would remove shared projects from resource. staged in common.go so that it can be used by other service connection but not doing the leg work for that at the moment due to scope.

xuzhang3 (changes requested):
@leasley199  Please help resolve the whitespace encoding issue. And can you help add the AccTest to cover the new changes(`shared_project_ids`) ?
image

leasley199:
@xuzhang3 can I get a review please, many thanks :)

xuzhang3 (changes requested):
@leasley199  we use '\t' as the white space to format the source code.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
```diff
@@ -263,8 +303,32 @@ func resourceServiceEndpointAzureRMUpdate(d *schema.ResourceData, m interface{})
 			return err
 		}
 	}
-	_, err = updateServiceEndpoint(clients, serviceEndpoint)
-
+    // Handle shared_project_ids updates
+    if d.HasChange("shared_project_ids") {
+        old, new := d.GetChange("shared_project_ids")
+        oldSet := toStringSet(old.([]interface{}))
```
xuzhang3: The built-in function `schema.NewSet(f SchemaSetFunc, items []interface{}) *Set` can be used to process Set collections.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
 	features = append(features, feature)
 	return features
 }
+
+func TestAccAzureRMServiceEndpointSharedProjectIDs(t *testing.T) {
```
xuzhang3: This file contains only unit tests.

Please put AccTest at https://github.com/microsoft/terraform-provider-azuredevops/blob/main/azuredevops/internal/acceptancetests/resource_serviceendpoint_azurerm_test.go

xuzhang3 (changes requested):

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
leasley199: updated, hopefully this fits the standard now.

xuzhang3: There are two different `_test.go` files, one if for unit test and another is for acctest.  All the `_azurerm` acctests are located at [resource_serviceendpoint_azurerm_test.go](https://github.com/microsoft/terraform-provider-azuredevops/blob/main/azuredevops/internal/acceptancetests/resource_serviceendpoint_azurerm_test.go). Since this test is an AccTest, we should add put this test at the AccTest file not the unit test.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm.go
```diff
@@ -546,3 +620,33 @@ func shouldValidate(features map[string]interface{}) bool {
 	}
 	return validate
 }
+
+
+func validateProjectID(d *schema.ResourceData, projectID string, m interface{}) (*uuid.UUID, error) {
```
xuzhang3: All the service connection sharing the same logic to set the project ID at [doBaseFlattening](https://github.com/microsoft/terraform-provider-azuredevops/blob/main/azuredevops/internal/service/serviceendpoint/commons.go#L314), `validateProjectID` can be removed as it do duplicate work

xuzhang3 (changes requested):

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
leasley199: ahhh my bad, read comment wrong thought you want them re-write as acc-test not a different file. Add back in unit-test and added accTest to accTest file. Thanks :)

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
 	features = append(features, feature)
 	return features
 }
+
```
xuzhang3, line 568: These tests can be remove since the AccTest has be relocated to AccTest file

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
...
+func TestAzureRMServiceEndpointSharedProjectIDs(t *testing.T) {
+    mainProjectID := uuid.New().String()
+    sharedProjectID1 := uuid.New().String()
+    sharedProjectID2 := uuid.New().String()
+
+
```
xuzhang3, line 573:
```suggestion
        resourceType := "azuredevops_serviceendpoint_azurerm"
	tfSvcEpNode := resourceType + ".test"
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
...
+	resource.Test(t, resource.TestCase{
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: func(s *terraform.State) error {
+			buildClient := azdosdkmocks.NewMockServiceendpointClient(gomock.NewController(t))
+			return testAccCheckAzureRMServiceEndpointDestroy(s, buildClient)
+		},
```
xuzhang3:
```suggestion
		testutils.CheckServiceEndpointDestroyed(resourceType),
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
 	features = append(features, feature)
 	return features
 }
+
+func TestAzureRMServiceEndpointSharedProjectIDs(t *testing.T) {
+    mainProjectID := uuid.New().String()
+    sharedProjectID1 := uuid.New().String()
+    sharedProjectID2 := uuid.New().String()
```
xuzhang3: These project should be the real project in the HCL, not dummy values
Example:
```hcl
func hclAzureRMServiceEndpointEnvironmentAzureShareProject(serviceEndpointName string) string {
	return fmt.Sprintf(`
resource "azuredevops_project" "test" {
  name = "%[1]s"
}

resource "azuredevops_project" "share" {
  name = "%[1]sshare"
}

resource "azuredevops_serviceendpoint_azurerm" "test" {
  project_id                        = azuredevops_project.test.id
  service_endpoint_name = "%[1]s"
  azurerm_subscription_id           = "00000000-0000-0000-0000-000000000000"
  azurerm_subscription_name         = "Test Subscription"
  azurerm_spn_tenantid              = "00000000-0000-0000-0000-000000000000"
  service_endpoint_authentication_scheme = "ServicePrincipal"
  credentials {
	serviceprincipalid  = "00000000-0000-0000-0000-000000000000"
	serviceprincipalkey = "serviceprincipalkey"
  }
  shared_project_ids = [azuredevops_project.share.id]
}`, serviceEndpointName)
}
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
...
+  azurerm_spn_tenantid              = "00000000-0000-0000-0000-000000000000"
+  service_endpoint_authentication_scheme = "ServicePrincipal"
+  credentials {
+    serviceprincipalid  = "e31eaaac-47da-4156-b433-9b0538c94b7e"
+    serviceprincipalkey = "serviceprincipalkey"
+  }
+  shared_project_ids = [%s]
+}
+`, mainProjectID, sharedProjectIDsConfig)
+}
```
xuzhang3:
```suggestion
func hclAzureRMServiceEndpointEnvironmentAzureShareProject(serviceEndpointName string) string {
	return fmt.Sprintf(`
resource "azuredevops_project" "test" {
  name = "%[1]s"
}

resource "azuredevops_project" "share" {
  name = "%[1]sshare"
}

resource "azuredevops_serviceendpoint_azurerm" "test" {
  project_id                        = azuredevops_project.test.id
  service_endpoint_name = "%[1]s"
  azurerm_subscription_id           = "00000000-0000-0000-0000-000000000000"
  azurerm_subscription_name         = "Test Subscription"
  azurerm_spn_tenantid              = "00000000-0000-0000-0000-000000000000"
  service_endpoint_authentication_scheme = "ServicePrincipal"
  credentials {
	serviceprincipalid  = "e31eaaac-47da-4156-b433-9b0538c94b7e"
	serviceprincipalkey = "serviceprincipalkey"
  }
  shared_project_ids = [azuredevops_project.share.id]
}`, serviceEndpointName)
}
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_azurerm_test.go
```diff
@@ -561,3 +565,96 @@ func initializeFeaturesWithValidate(validate bool) []map[string]interface{} {
...
+
+func testCheckAzureRMServiceEndpointDestroy(s *terraform.State, buildClient serviceendpoint.Client) error {
```
xuzhang3, line 631: You can use `testutils.CheckServiceEndpointDestroyed(tfNode)` instead.

xuzhang3 (changes requested):
Please ensure the code can be complied and the tests passed, these are our criteria for merging PRs

Issue #615, opened 2022-07-05
Adding Secure File using terraform


rajeshyadavk2:
Hi,

How can we add the secure file using terraform(Library-->Securefiles)? Can you please suggest me a way to add the id_rsa file to the azure devops using terraform code? 
I'm stuck at this point. Any help will be highly appreciated!!

Thanks in advance
Rajesh beri

xuzhang3:
@rajeshyadavk2  are you try to read the `id_isa` file when running terraform? you export the file content as ENV variable.

rajeshyadavk2:
> @rajeshyadavk2 are you try to read the `id_isa` file when running terraform? you export the file content as ENV variable.

I'm trying to upload the id_rsa file using terraform code. 
I will use this id_rsa file inside the task(**Install an SSH key**) 

# Install SSH key
# Install an SSH key prior to a build or deployment
- task: InstallSSHKey@0
  inputs:
    knownHostsEntry: 
    sshPublicKey: 
    #sshPassphrase: # Optional
    sshKeySecureFile: 
    #addEntryToConfig: # Optional
    #configHostAlias: # Optional
    #configHostname: # Optional
    #configUser: # Optional
    #configPort: # Optional
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/install-ssh-key?view=azure-devops


Because i'm not getting any way to do it till now.
Can you suggest any other way to this via terraform??

xuzhang3:
@rajeshyadavk2  manage secure files is not supported by this provider and the API is not correctly documented.
 https://www.terraform.io/language/resources/provisioners/local-exec may be the alternative choice for you

Issue #274, opened 2021-01-21
Possible to add secure files to the Pipeline Library?


CaptainStealthy:
Is it possible to add secure files to the Library using this provider?

Looked through the docs, but didn't find anything.

Thanks!

![securefiles](https://user-images.githubusercontent.com/48744965/105427080-99d0ef80-5c09-11eb-8040-338d33822e0b.png)

xuzhang3:
Hi @CaptainStealthy  Seem cannot implement this resource for ADO provider because there's no secure file APIs.

tmegow:
I am terraforming Azure Devops Pipelines which use Secure Files and I also desire this feature.

CaptainStealthy:
What kind of secure files are you trying to store? We were using it for certificates, but we worked around the issue by storing them in Azure Key Vault.

Even if you're not storing certs, you can base64 encode your files and store them as Key Vault Secrets, maybe?

KjeldSchmidt:
We are still interested in this, for what it's worth 🙂

ianhockett:
FWIW it appears there is an API for secure files, but it is absent from the documentation. [Here](https://github.com/MicrosoftDocs/vsts-rest-api-specs/issues/623) is an open issue for the API docs.

We are able to upload secure files via REST call using this uri: `..._apis/distributedtask/securefiles?api-version=7.1-preview.1...`.

Phiph:
I'd be interested in this

jubr:
Hi @xuzhang3, did you happen to see @ianhockett's comment that the API in question exists?   
I would also be very interested in having being able to manage Secure Files with the terraform Azure Devops provider!

@ others: please add a 👍 to the original post to indicate interest, thx!

xuzhang3:
@jubr  azuredevops go-sdk not include all the APIs, some of the APIs are missing in the SDK. Sometimes we add APIs manually to supplement the SDK.

magodo:
This issue is blocked by missing of the API definition, as well as the corresponding Go SDK. See: https://github.com/MicrosoftDocs/vsts-rest-api-specs/issues/623.

PR #1380, opened 2025-05-30
Added secure file integration with resource authorization.

aklakina:
# Secure file integration

These changes are adding the secure file resource and data integration to azure devops (ADO going forward).

The API is not implemented in the ADO go API, therefore I had to use the raw ADO API to send REST API requests to custom location.

Secure files in ADO have a GUID constructed ID, yet their name must be unique too in the same project library. This could lead to conflicts, where the terraform can no longer read the resource because of 404 (assuming the previous secure file got recreated) and yet it won't be able to deploy either, because the name is already taken. Therefore this module uses name query to refresh the state of a secure file. This makes the content of the secure file non-write-only, although this leads to an issue where we would not know if the contents are the same. Thus this module introduces the file_hash_sha1 and 256 properties, which are automatically added to the secure file entity in ADO. Then the difference can be calculated 2 ways:
1. Checking the state file for content change
2. Reading the file_hash_sha1 and 256 respectively and comparing the hash for the current content in the configuration.

The content can not be updated in-place, so these properties are marked as `ForceNew`.

The resource authorization endpoint were the same as it was for the variable group, but instead of `variablegroup` url part, it needed the `securefile` url part.
The permission resource works the same way as it is working for the variable group.

## All Submissions:

Issue Number:
- #615 
- #274

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


aklakina:
Can I please get an ETA for a review on this PR? @xuzhang3

kmaikaze:
What has to happened to put more attention on this PR?
@xuzhang3

aklakina:
Hi @magodo

Can you please take a look at this pr?

magodo:
@aklakina Would you please add some acceptance tests to cover the new resources and data source?

aklakina:
I'll be able to add some tests next week

aklakina:
@magodo

I have added acceptance tests for the resource and data source implementations.

azuredevops/internal/service/taskagent/resource_secure_file.go
```diff
@@ -0,0 +1,472 @@
...
+				Computed:    true,
+				ForceNew:    true,
+				Description: "SHA1 hash of the file content. Computed from the content during creation.",
+			},
+			"file_hash_sha256": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "SHA256 hash of the file content. Computed from the content during creation.",
+			},
```
magodo, line 82: Any reason to introduce twos? Also computed only attributes shall not specify `ForceNew`.

azuredevops/internal/service/taskagent/resource_secure_file.go
```diff
@@ -0,0 +1,472 @@
...
+		},
+	}
+}
+
+// getSecureFileURL builds the secure file URL for a given HTTP method.
+func getSecureFileURL(clients *client.AggregatedClient, projectID, secureFileID string, params url.Values) string {
+	base := strings.TrimRight(clients.OrganizationURL, "/") + "/" + strings.TrimLeft(projectID, "/") + "/_apis/distributedtask/securefiles"
+	baseParams := url.Values{}
+	baseParams.Add("api-version", "6.0-preview.1")
```
magodo, line 91: This API version isn't even available in the ADO api reference, though it is available in the API spec..

azuredevops/internal/service/taskagent/resource_secure_file.go
```diff
@@ -0,0 +1,472 @@
...
+				ForceNew:    true,
+				Description: "SHA256 hash of the file content. Computed from the content during creation.",
+			},
+		},
+	}
+}
+
+// getSecureFileURL builds the secure file URL for a given HTTP method.
+func getSecureFileURL(clients *client.AggregatedClient, projectID, secureFileID string, params url.Values) string {
+	base := strings.TrimRight(clients.OrganizationURL, "/") + "/" + strings.TrimLeft(projectID, "/") + "/_apis/distributedtask/securefiles"
```
magodo, line 89: Can you share some context about the API definition for this endpoint? I can't find it anywhere, including the API spec: https://github.com/MicrosoftDocs/vsts-rest-api-specs/blob/master/specification/distributedTask/6.1/taskAgent.json

azuredevops/internal/service/taskagent/resource_secure_file.go
aklakina: I've added two to reduce the probability of 2 distinct file contents being considered as if they were the same. sha256 already has a pretty low probability though so we can remove sha1 if you wish.

aklakina: As for the `ForceNew`, it is necessary here because the implementation is using file names as ids instead of the guid assigned to it by ADO. This way if a secure file is replaced on the cloud but not in the tfstate you wouldn't know since the content can not be queried. If the hashes change, it has to replace the file, otherwise you'll have a malformed state.

azuredevops/internal/service/taskagent/resource_secure_file.go
aklakina: I didn't find it anywhere either in docs. I had to watch the network traffic to identify the endpoint and how it behaves. From that I used postman to figure out what could be done with it and what not.

azuredevops/internal/service/taskagent/resource_secure_file.go
aklakina: Without the `-preview.1` the endpoint throws
```
{"$id":"1","innerException":null,"message":"The requested version \"6.0\" of the resource is under preview. The -preview flag must be supplied in the api-version for such requests. For example: \"6.0-preview\"","typeName":"Microsoft.VisualStudio.Services.WebApi.VssInvalidPreviewVersionException, Microsoft.VisualStudio.Services.WebApi","typeKey":"VssInvalidPreviewVersionException","errorCode":0,"eventId":3000}
```
And the webui is using this `-preview.1` too based on the network traffic.

aklakina:
@magodo

any update on this one?

magodo:
@aklakina Sorry for the delay. My main cocern about this PR is using the undocumented API, which has no stability guarantee.

aklakina:
Sadly I don't have the power to change on this and I did not see any other API that manages secure files. So what do you suggest?

magodo:
Unfortunately for this one, I'd be refrained to get this in to avoid unnecessary breaking changes from the API.

aklakina:
Well, if the endpoint has been documented properly and the sdk has been extended with the secure files API, you can drop me a ping in this pr and I'll change the things needed.

aklakina:
@magodo

As an extra information, I found this [issue](https://github.com/MicrosoftDocs/vsts-rest-api-specs/issues/623) on the docs repository. Do you think you could push this topic internally? It would be really helpful for everybody I think.

magodo:
@jaredfholgate Do you have more context about this? Especially the issue https://github.com/MicrosoftDocs/vsts-rest-api-specs/issues/623 mentioned above.

aklakina:
Hi @magodo, @jaredfholgate 

Any news with the documentation of this API endpoint?
The API endpoint is even versioned and is used openly by the frontend.

jaredfholgate:
> Hi @magodo, @jaredfholgate
> 
> Any news with the documentation of this API endpoint? The API endpoint is even versioned and is used openly by the frontend.

Hi. Apologies for the lack of progress on this. @magodo is in an active discussion with the Azure DevOps product group on this.

Unfortunately, it is not straight forward to implement due to the way the APIs work and eventual consistency. It doesn't play nicely with a standard Terraform provider implementation. Hopefully there will be a work around available for this provider soon.

aklakina:
Dropping a ping here to see if there is any news about the progress here
@magodo @jaredfholgate

PR #1396, opened 2025-06-27
Move the PR CI from ADO to github action

magodo:
There is an internal ADO pipeline for CI check, which is not triggered automatially (not sure why). Instead, we can simply use github action for the PR CI check.

This PR adds all the necessary github actions for the PR CI check:
- depscheck
- go-lint (with all lint errors fixed)
- terrafmt
- test (UT and gofmt check)


.golangci.yml
```diff
@@ -1,39 +1,90 @@
...
+
+    ###### DISABLED because golang will put the space back into `//nolint: linter` -> `// nolint: linter` ######
+    #- nolintlint  #Reports ill-formed or insufficient nolint directives.
+    
+    ###### DISABLED because no easy way to format tags
+    #- tagalign #  checks struct tags that do not align with the specified column in struct definitions.
 
 linters-settings:
   errcheck:
-    check-blank: true
```
ziyeqf, line 31: we don't need this anymore?

ziyeqf (approved):

Issue #1385, opened 2025-06-04
azuredevops_serviceendpoint_azurerm / remove subscription scope dependency


keisari-ch:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When creating an azurerm service connection (automatic or manual, with the GUI or TF), a subscription (or management group) needs to be passed to the resource.

Using the GUI, there's an automatic Owner role assignement on the subscription scope.

When i create the app reg separately (manual), the subscription information is still required, but no role binding is made.

What if i just want an application on the Azure side, and with the associated permissions, i need to either :

- Manage entra groups only.
- Or Manage applications only
- Or have a service principal used by terraform to only manage network with a custom role on a set of subscriptions
- Or have a service principal used by terraform to only manage DNS Zones across all the resource groups.

Basically, i have a bunch of use cases where i need to remove the default Owner role assignement. For the sake of least privilege.

I'm actually quite surprised of the default behavior, assigning the Contributor role by default.

### New or Affected Resource(s)

* azuredevops_serviceendpoint_azurerm

### Potential Terraform Configuration

```hcl
resource "azuredevops_serviceendpoint_azurerm" "example" {
  project_id                             = azuredevops_project.example.id
  service_endpoint_name                  = local.service_connection_name
  description                            = "Managed by Terraform"
  service_endpoint_authentication_scheme = "WorkloadIdentityFederation"
  credentials {
    serviceprincipalid = azurerm_user_assigned_identity.example.client_id
    # or  serviceprincipalid = azuread_application_federated_identity_credential.example.client_id
  }
  azurerm_spn_tenantid      = "00000000-0000-0000-0000-000000000000"
  # azurerm_subscription_id   = "00000000-0000-0000-0000-000000000000"  # removed
  # azurerm_subscription_name = "Example Subscription Name"             # removed
}

```

### References

* #0000

Issue #1367, opened 2025-05-08
Add new resource for aqua platform service connection/ednpoint


khamidziyo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### New or Affected Resource(s)

* azuredevops_serviceendpoint_aquaplatform

### Potential Terraform Configuration

```hcl
data "azuredevops_serviceendpoint_aquaplatform" "example" {
    project_id = data.azuredevops_project.example.id
    name = "my-trivy-aqua-platform-connection"
}

resource "azuredevops_serviceendpoint_aquaplatform" "aquasec" {
  project_id            = data.azuredevops_project.example.id
  aqua_platform_url            = var.aquasec_portal_server_url
  aqua_auth_url              = "https://api.cloudsploit.com"
  aqua_key = var.aqua_key
  aqua_secret = var.aqua_secret
  service_endpoint_name = "my-trivy-aqua-platform-connection"
  description           = "AquaSec Portal service connection for Trivy task"
}


```

### References

Extension in Marketplace: https://marketplace.visualstudio.com/items?itemName=AquaSecurityOfficial.trivy-official
Configuration documentation: https://github.com/aquasecurity/trivy-azure-pipelines-task/blob/main/docs/connectedservice.md

PR #1232, opened 2024-11-20
Feature: Add JFrog V2 Service Endpoint Data Source (Issue #1231)

jasonouellet:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [ ] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## What about the current behavior has changed?

Issue Number: #1231 

Ading data source:

* azuredevops_serviceendpoint_jfrog_artifactory_v2
* azuredevops_serviceendpoint_jfrog_distribution_v2
* azuredevops_serviceendpoint_jfrog_platform_v2
* azuredevops_serviceendpoint_jfrog_xray_v2

## Does this introduce a change to `go.mod`, `go.sum` or `vendor/`?

- [ ] Yes
- [x] No

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Any relevant logs, error output, etc?

(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Issue #1231, opened 2024-11-20
Data Source - azuredevops_serviceendpoint_jfrog_[*]_v2 - Missing


jasonouellet:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As the title says, the data sources are missing. I think it is quite similar to the other existing service endpoint.

### New or Affected Resource(s)

* azuredevops_serviceendpoint_jfrog_artifactory_v2
* azuredevops_serviceendpoint_jfrog_distribution_v2
* azuredevops_serviceendpoint_jfrog_platform_v2
* azuredevops_serviceendpoint_jfrog_xray_v2

### Potential Terraform Configuration

```hcl
data "azuredevops_serviceendpoint_jfrog_[*]_v2" "example" {
  project_id            = data.azuredevops_project.example.id
  service_endpoint_name = "Example JFrog [*] V2"
}
```

### References

* N/A

PR #1400, opened 2025-07-04
New resource: `azuredevops_serviceendpoint_generic_v2`

aklakina:
# Disclaimer
I know there are lot of pull requests and issues trying to cover this issue but I didn't find any that is generic enough to be useable by any and all service endpoint integration.

# Description
This pull request introduces comprehensive support for managing and retrieving Service Endpoints through Terraform in a standardized manner.

The following key changes are included:

## 1. New Resource: `azuredevops_serviceendpoint_generic_v2`

- **Purpose:**  
  Enables users to create and manage Service Endpoints in Azure DevOps, supporting integration with external services using custom authentication mechanisms.
- **Features:**  
  - Directly integrates with how Azure Devops manages service endpoints
  - Has minimal amount of required attributes, everything else is gathered from Azure Devops through api.
  - The configuration is validated during plan time to evaluate what data is missing or not appropriate.
- **Documentation:**  
  - Provides detailed usage examples for different authentication types.
  - Lists all supported arguments and exported attributes.
  - Includes import instructions for existing service endpoints.

## 2. New Data Source: `azuredevops_serviceendpoint_generic_v2`

- **Purpose:**  
  Allows users to retrieve information about existing Service Endpoints by either ID or name, facilitating data-driven Terraform configurations.
- **Features:**  
  - Supports lookup by both `service_endpoint_id` and `service_endpoint_name`.
  - Exposes all relevant attributes, including authorization details and custom data.
- **Documentation:**  
  - Offers example queries for both lookup methods.
  - Clearly documents all arguments and exported attributes.

## 3. Documentation

- **Resource Documentation:**  
  - Located in `website/docs/r/serviceendpoint_generic_v2.html.markdown`.
  - Explains resource arguments, attributes, and provides multiple usage scenarios.
- **Data Source Documentation:**  
  - Located in `website/docs/d/serviceendpoint_generic_v2.html.markdown`.
  - Details data source arguments, attributes, and example queries.

## 4. Benefits

**Standardized**:

No need for specific service endpoint resource or data besides this.

---

These enhancements significantly improve the Azure DevOps Terraform provider by enabling any service endpoint scenario without the need for adding a new service endpoint integration.

## All Submissions:

* [ ] My code follows the code style of this project.

Issue Number:
- #219
- #628
- #683
- #767
- #926
- #1367 
- #1231 and #1232
- #1385
- #1376

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


aklakina:
Can you please review this pr? I understand that this is a major change in how you handle the service endpoints but this would be very useful as you wouldn't need distinct resources and bunch of new code just for the sake of a service endpoint integration.
@xuzhang3

website/docs/r/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Required) The ID of the project to which the service endpoint belongs.
+* `service_endpoint_name` - (Required) The name of the service endpoint.
```
magodo: This can be shortened to `name`?

website/docs/r/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+* `service_endpoint_type` - (Required) The type of the service endpoint. This can be any valid service endpoint type, such as "generic", "artifactory", etc.
```
magodo: This can be shortened to `type`?

website/docs/r/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,79 @@
...
+* `description` - (Optional) The description of the service endpoint. Defaults to "Managed by Terraform".
+* `server_url` - (Required) The URL of the server associated with the service endpoint.
+* `authorization_type` - (Required) The authorization scheme to use. Common values include "UsernamePassword", "Token", "OAuth", etc.
```
magodo: Shall we rename this to `authorization_scheme` to make it consistent with the API reference?

xuzhang3:
> Can you please review this pr? I understand that this is a major change in how you handle the service endpoints but this would be very useful as you wouldn't need distinct resources and bunch of new code just for the sake of a service endpoint integration. @xuzhang3

Sorry, no longer working at MS.

aklakina:
Hi @magodo

Can you please take a look at this pr?

aklakina:
I've also implemented the changes from #1351

Couldn't test them yet, will do in the following week.

main.go
```diff
@@ -3,7 +3,6 @@ package main
 import (
```
magodo, line 3: This change seems not related to this PR, can we revert it?

website/docs/d/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+* `project_id` - (Required) The ID of the project to which the service endpoint belongs.
+* `service_endpoint_id` - (Optional) The ID of the service endpoint to retrieve. One of `service_endpoint_id` or `service_endpoint_name` must be specified.
+* `service_endpoint_name` - (Optional) The name of the service endpoint to retrieve. One of `service_endpoint_id` or `service_endpoint_name` must be specified.
+
+## Attributes Reference
+
+In addition to the arguments listed above, the following attributes are exported:
+
+* `id` - The ID of the service endpoint.
+* `service_endpoint_type` - The type of the service endpoint.
```
magodo: Same here

website/docs/d/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+* `description` - The description of the service endpoint.
+* `server_url` - The URL of the server associated with the service endpoint.
+* `authorization_type` - The authorization type of the service endpoint.
```
magodo: Same here

website/docs/d/serviceendpoint_generic_v2.html.markdown
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `project_id` - (Required) The ID of the project to which the service endpoint belongs.
+* `service_endpoint_id` - (Optional) The ID of the service endpoint to retrieve. One of `service_endpoint_id` or `service_endpoint_name` must be specified.
+* `service_endpoint_name` - (Optional) The name of the service endpoint to retrieve. One of `service_endpoint_id` or `service_endpoint_name` must be specified.
```
magodo: Same here

azuredevops/internal/acceptancetests/resource_serviceendpoint_generic_v2_test.go
```diff
@@ -0,0 +1,366 @@
+//go:build (all || resource_serviceendpoint_generic_v2) && !exclude_resource_serviceendpoint_generic_v2
+// +build all resource_serviceendpoint_generic_v2
+// +build !exclude_resource_serviceendpoint_generic_v2
```
magodo: Can we remove these?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
```diff
@@ -0,0 +1,818 @@
...
+		Authorization: &serviceendpoint.EndpointAuthorization{
+			Scheme:     converter.String(config.AuthType),
+			Parameters: &config.AuthData,
+		},
+	}
+
+	// Handle additional data
+	if len(config.Data) > 0 {
+		serviceEndpoint.Data = &config.Data
+	}
```
magodo, line 419: Based on the API reference, this seems to only exist in the response?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
```diff
@@ -0,0 +1,818 @@
...
+	if serviceEndpoint == nil {
+		d.SetId("")
+		return nil
+	}
+
+	return updateResourceDataFromServiceEndpoint(d, serviceEndpoint)
+}
+
+// updateResourceDataFromServiceEndpoint updates the resource data with values from the service endpoint
+func updateResourceDataFromServiceEndpoint(d *schema.ResourceData, serviceEndpoint *serviceendpoint.ServiceEndpoint) diag.Diagnostics {
```
magodo: Can we inline this to the call site?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
```diff
@@ -0,0 +1,818 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid service endpoint ID: %w", err)
+	}
+
+	args := serviceendpoint.DeleteServiceEndpointArgs{
+		EndpointId: &endpointUUID,
+		ProjectIds: &projectIDs,
+	}
+
+	err = clients.ServiceEndpointClient.DeleteServiceEndpoint(ctx, args)
```
magodo, line 777: Will the delete call also delete the shared references automatically?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
```diff
@@ -0,0 +1,818 @@
...
+func validateServiceEndpointSchema(clients *client.AggregatedClient, serviceEndpoint EndpointConfig, planTime bool) error {
+	serviceEndpointType := serviceEndpoint.ServiceEndpointType
+
+	// Check if types have been initialized yet
+	serviceEndpointTypesMutex.RLock()
+	initialized := serviceEndpointTypesInitialized
+	serviceEndpointTypesMutex.RUnlock()
+
+	// If not initialized, initialize the cache
+	if !initialized {
```
magodo: There will be a race condition if multiple service endpoint resources are provisioning in parallel.

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: Maybe a better place for this would be in the provider initialization anyways. Then there is no need for mutex, cause it only ever happens once per provider.

aklakina:
Will fix next week

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: The Data property in the API refers to extension input descriptor outside of authentication schema:
https://learn.microsoft.com/en-us/azure/devops/extend/develop/service-endpoints?view=azure-devops
Here you can see that inputdescriptor property can be added without the parent property authenticationschema. This will be represented as the `data` in the API, which also can be required or empty.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: `DeleteServiceEndpointArgs` API takes in a list of project ids as reference. So unless the go api behaves unexpectedly, I would say yes.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: BTW this shouldn't create race condition as the `InitServiceEndpointTypes` uses exclusive mutex lock on `serviceEndpointTypesMutex` to initialize the types. So what happens is that the processes check if it is initialized with a read lock, if not, then it will call the init func which uses write lock so 1 process enters, rest will wait. Then it inits the types and sets the `serviceEndpointTypesInitialized` to true. Next process comes from mutex queue and the first statement:
```go
if serviceEndpointTypesInitialized {
	return nil // Already initialized
}
```
Will already cause the func to exit.

Maybe it would still make sense to move this out to provider initialization, but that needs intrusive modifications.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_generic_v2.go
```diff
@@ -0,0 +1,194 @@
...
+				Optional:     true,
+				ValidateFunc: validation.IsUUID,
+				ExactlyOneOf: []string{"id", "name"},
+			},
+			// Use name to look up by name
+			"name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+				ExactlyOneOf: []string{"id", "name"},
```
magodo, line 50: These two attributes need to be Optional+Computed, so that when "A" is specified, "B" is still exposed to be used by the user.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
magodo: In this case you shall use `sync.Once`.

magodo (changes requested):

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_generic_v2.go
aklakina: I've added the Computed: true schema attribute.

website/docs/r/serviceendpoint_generic_v2.html.markdown
aklakina: Renamed in docs and all other places.

website/docs/r/serviceendpoint_generic_v2.html.markdown
aklakina: Renamed in docs and all other places.

website/docs/r/serviceendpoint_generic_v2.html.markdown
aklakina: Renamed in docs and all other places.

main.go
aklakina: I've reverted the changes on main.go. Originally I did them because the linter was complaining.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: Moved it to be inline in the calling func.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_generic_v2.go
aklakina: Modified it so it uses sync.once.

magodo:
@aklakina Thanks for the update! Would you please resolve the CI failures and attach a screenshot/command line paste to show the acceptance tests are passing? Then I'll be happy to merge this PR :+1:

aklakina:
It needed some fixes, but it should be fine now.
image

magodo:
@aklakina Thank you for updating the test result! Would you please fix the CI failure, then I'll be happy to merge it.

aklakina:
I fixed the CI failure probably now. Used terrafmt to fmt the terraform blocks.

magodo (approved):

aklakina:
Hi @magodo

Would you agree to create a new pr that adds a deprecation message to all type-specific service endpoint resource, saying that one should use this resource instead?

magodo:
@aklakina I'm refrained to do so.

Issue #1366, opened 2025-05-07
Feature: Authentication - Support Azure DevOps ID Token Refresh


jaredfholgate:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support ID Token Refresh when using service principal authentication with OIDC.

Here is the equivalent AzAPI implementation for reference:

- Service Connection ID Input: 
    - https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L66
    - https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L283
    - https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L529

- SYSTEM_ACCESSTOKEN support: https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L504

- Build credentials: 
  - https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L809
  - https://github.com/Azure/terraform-provider-azapi/blob/318f525558caa76b738b80f917c82eb3b253022b/internal/provider/provider.go#L962

- Use the Pipelines Credentials package: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/azure_pipelines_credential.go
  - NOTE the hard coding of the env var for the url:  https://github.com/Azure/azure-sdk-for-go/blob/9e8eb1f00372627ba42158690e604ccbcba11bf6/sdk/azidentity/azure_pipelines_credential.go#L22

### New or Affected Resource(s)

N/A, provider scoped

### Potential Terraform Configuration

N/A

### References

N/A

michvllni:
Is there any update on this? We direly need this. At the moment we have to precisely watch our pipelines to ensure the token does not time out betwen plan and apply (different jobs with an approval gate)

[...]

Issue #419, opened 2021-07-14
Allow authentication in same manner as azurerm via azure cli login


haymansfield:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I can create azure devops resources using the azure cli azure-devops extension without the need for managing PAT tokens. Could this provider be made to use the same authentication?

### New or Affected Resource(s)

* azuredevops_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

tmeckel:
Problem here is that the provider may work longer on a deployment than the life time of the access token. In that case the provider can use a refresh token to request a new access token. The problem is that we don't have anything in place right now to handle an expired login/access token, because the provider works with a (very) long living PAT. So implementing this would be a major effort. To make the situation even worse, the code that handles the authentication is hosted in a different repository https://github.com/microsoft/azure-devops-go-api.

I successfully tested the [OAuth 2.0 Resource Owner Password Credentials (ROPC) grant](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth-ropc) and Azure DevOps, where you use an application in Azure AD which impersonates a real user to access Azure DevOps. This flow is completely non-interactive and thus suitable for example CI/CD scenarios. The only drawback is that you need both the application secret **and** the passoword of the impersonated user.

![image](https://user-images.githubusercontent.com/14177833/126908067-6a6995dd-61d3-41db-8aed-b94b3f3b0e61.png)

> Hint: You always need a real user object while accessing Azure DevOps because Azure DevOps does not support Azure AD service principals.

tmeckel:
[...]

PR #1402, opened 2025-07-11
Replace ADO sdk by a fork and support AAD based auth token refresh

### LINKED ISSUE #1176: [FEAT]: `skip_provider_authentication` provider option
### Community Note

### Description

I appreciate this is something of an edge case, but it would help to significantly reduce the amount of code we have to managed in our Azure Landing Zones and Azure Verified Modules.

We support both Azure DevOps an GitHub as options for our customers to bootstrap their version control system.

Due to a limitation of Terraform Core, we are unable to reference both the `azuredevops` and `github` provider in the same module. Terraform Core does not support optional providers. It always attempts to initialise the provider even if no resources or data sources for that provider are part of the graph.

If someone chooses the GitHub option, then we cannot expect them to supply working Azure DevOps credentials as they don't even have access to Azure DevOps.

This means that every module in our module tree has to have 3 versions: `azuredevops`, `github` and `local` (local file system). This is a significant maintenance overhead requiring a lot of copy and paste code.

In order to work around this it would be great if we could add a flag that skips attempting to auth the provider and just returns a success from the init command.

e.g.

```hcl
provider "azuredevops" {
   skip_provider_authentication = var.version_control_system == "github"
}
```

The obvious limitation with this is that if someone does reference an Azure DevOps resource or data source when this provider variable is set they will get an auth error.

If this is something you think could work I would be happy to submit a PR for the code. 

This is a provider variable.

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

magodo:
This PR replaces the go module `github.com/microsoft/azure-devops-go-api/azuredevops/v7` by fork `github.com/magodo/azure-devops-go-api/azuredevops/v7`, which provides a way to allow the client get a refreshed token when expired.

Fix #1176, 

Relating to:
- #1366
- #419


magodo:
@tmeckel Appreciate if you can take a look at this PR.

ziyeqf (approved):

Issue #203, opened 2020-10-13
Add support for service hooks


richardgavel:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add support to maintain service hooks (specifically webhooks)

### New or Affected Resource(s)

* azuredevops_webhook

### Potential Terraform Configuration

```hcl
```

xuzhang3:
[REST API](https://docs.microsoft.com/en-us/rest/api/azure/devops/hooks/?view=azure-devops-rest-5.1)

richardgavel:
I can see why this hasn't been done yet. All the various kinds of triggers each require a potentially different block (buildCompleted vs codePushed vs etc). I'm starting to put it together myself, using serviceendpoint as a model

```
azuredevops_servicehook_webhook XXX
{
  trigger = "codePushed"

  codePushedTrigger = {
    repository = XXX
    branchName = XXX
    pushedBy = XXX
  }

  action = "HttpPost"

  httpPost ={
    url = ""
    etc..
  }
}
```

The trigger and the potential trigger block settings would be in the base, the other properties unique to the action would be in the each of the resource_servicehook_XXX resources (webhook vs slack vs etc, though I am only going to try and do webhook).

stobias123:
Bump! Really need this feature!

jamiejackson:
Did you get anywhere with it, @richardgavel ?

fernandosalomao:
https://github.com/microsoft/azure-devops-go-api/blob/dev/azuredevops/servicehooks/client.go

xuzhang3:
Supported resource: [azuredevops_servicehook_storage_queue_pipelines](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/servicehook_storage_queue_pipelines)

symnatbay:
Even just a service hook for a custom webhook target and its events would be very helpful for managing custom integrations in large organizations.

magodo:
Fixed by #1412

symnatbay:
[...]

PR #1412, opened 2025-08-07
New resource: `azuredevops_servicehook_webhook_tfs`

dawwestk:
# Description

This pull request introduces a new Servicehook Webhook TFS resource type. It also consolidates the Servicehook logic with the existing `azuredevops_servicehook_storage_queue_pipelines` so that addition of future Servicehook types is simplified.

The resource name follows the existing Servicehook resource convention of `consumer` followed by event `publisher`.

The changes are as follows:

## 1. New Resource: `azuredevops_servicehook_webhook_tfs`

**Purpose:** To allow creation of [Webhook type](https://learn.microsoft.com/en-us/azure/devops/service-hooks/consumers?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#webhooks) Servicehooks with any of the [TFS publisher events](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops).

**Features:** All possible inputs for Webhooks and TFS events are included.

## 2. Documentation

Resource documentation can be found in `website/docs/r/servicehook_webhook_tfs.html.markdown`

## 3. Bug Fix for `azuredevops_servicehook_storage_queue_pipelines` deletion

If a Servicehook of type Azure Storage is created by terraform, then manually deleted outside of terraform's control (i.e. through the UI or API), the next run of `terraform plan` or `terraform apply` will fail with a "resource not found" error, rather than gracefully handling the not found case.

This process has been updated to use the same process as the newly created resource.

## All Submissions:

Issue Number:

#203 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

Related to bug fix for existing `azuredevops_servicehook_storage_queue_pipelines` resource.

Steps to reproduce:
* Create resource using terraform
* Delete via Azure DevOps Portal
* Re-run `terraform plan` to see not found case throwing error below

```shell
12:44 # terraform apply
azuredevops_servicehook_storage_queue_pipelines.example: Refreshing state... [id=c5188170-1d43-4119-99ec-4d71e5a3ac5b]
╷
│ Error: No subscription could be found with id c5188170-1d43-4119-99ec-4d71e5a3ac5b.
│ 
│   with azuredevops_servicehook_storage_queue_pipelines.example,
│   on main.tf line 709, in resource "azuredevops_servicehook_storage_queue_pipelines" "example":
│  709: resource "azuredevops_servicehook_storage_queue_pipelines" "example" {
│ 
```


dawwestk:

dawwestk:
Hi @xuzhang3 , please can you take a look at this PR? Cheers

dawwestk:
Hi @magodo please can you take a look at this PR? Cheers

magodo:
@dawwestk Sorry for the delay...

I have some basic questions regarding this new resource and the existing storage queue pipelines resource: Both resources are named in the form of `<consumer_type>_<event_type>`, especially for the Azure Storage and Webhook consumer types, the document stated that they support all events, do we have to call out the `<event_type>` in the resource type explicitly? Furthermore, I noticed that not all event types are declared in the new resource, as well as the old one, where it only supports a few pipeline related events. Does this indicate the document is incorrect, or you were doing this on purpose?

Another thing is that you mentioned the existing `azuredevops_servicehook_storage_queue_pipelines` will fail to remove itself from state if it is removed at ADO. I assume the error: `No subscription could be found with id c5188170-1d43-4119-99ec-4d71e5a3ac5b.` is raised by the API, which indicates the `err` of the call below is not nil:

```
	subscription, err := getSubscription(clients, subscriptionId)
	if err != nil {
		return err
	}
```

This means the code you added to nil check the `subscription` will not be run, and the issue still remains.

The last thing is that your PR seems doesn't include the document of the new resource.

dawwestk:
Thanks for taking a look, please see below for my responses to your queries:

## Naming

The decision to explicitly name the resource `<consumer_type>_<event_type>` was largely due to the existing Azure Storage example. However, I do think it makes sense to keep the configuration specific to each event type to ensure that each resource is focused and has a well-defined scope.

## Event types 

All event types with `Publisher ID: tfs` in [the MS documentation](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?view=azure-devops&toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json) have been included in the `tfs_publisher.go`. If there are others not documented here please let me know. 

The event types associated with the Azure Storage resource (defined in `pipelines_publisher.go`) have not been updated as this PR is not concerned with them.

## Nil Check

The `getSubscription` function can return errors other than nil subscription - in fact, in the subscription-is-nil case (i.e. not found by the `utils.ResponseWasNotFound` function) the function returns `(nil, nil)`. Then the subsequent nil check is performed and `d.SetId("")` is run - this pattern is used elsewhere in the repo as a single step (check ID and set ID in resource data as "") but I didn't want a get method to change state.

## Documentation

Apologies, the new resource docs are attached now

magodo:
@dawwestk Sorry, I'm still confused about some inqueries..

## Event Types

E.g. There is only one build&release event type gets covered by your code:

```
	// Build and Release Events
	"build_completed": "build.complete",
```

While the document has listed a few:

[Build completed](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#build.complete)
[Release abandoned](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.release-abandoned-event)
[Release created](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.release-created-event)
[Release deployment approval completed](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.deployment-approval-completed-event)
[Release deployment approval pending](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.deployment-approval-pending-event)
[Release deployment completed](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.deployment-completed-event)
[Release deployment started](https://learn.microsoft.com/en-us/azure/devops/service-hooks/events?toc=%2Fazure%2Fdevops%2Fmarketplace-extensibility%2Ftoc.json&view=azure-devops#ms.azure-devops-release.deployment-started-event)

## Nil Check

If the error you described in the issue is about `err` is a wrapper around http status code 404, the code you added still won't take effect. Shouldn't it be something like:

```go
	if err != nil {
		if utils.ResponseWasNotFound(err) {
			d.SetId("")
			return nil
		}
		return fmt.Errorf(...)
	}

```

dawwestk:
@magodo no worries, I'll try to clarify - cheers again for your input.

Those event types listed in your comment (apart from `build_completed`) all use the `rm` publisher, not the `tfs` one. So they would be handled by an (as-yet non-existent) resource type as this one only covers the `tfs` publisher and `webhook` events.

For the nil check, I'll take a look later and test your change - the existing code works as desired, but I can update to stay within best practice guidance.

dawwestk:
Based on your feedback I have updated `getSubscription` to propagate the error to the calling functions, where the `utils.ResourceNotFound()` check is then run:

```
	if err != nil {
		if utils.ResponseWasNotFound(err) {
			log.Printf("[INFO] Service hook subscription not found. ID: %s", d.Id())
			d.SetId("")
			return nil
		}
		return err
	}
```

The `getSubscription` function remains a simple wrapper for the API call. The 404 no longer crashes the provide on test (i.e. create service hook via terraform, delete manually, re-run `terraform plan`)

azuredevops/internal/service/servicehook/commons.go
```diff
@@ -0,0 +1,135 @@
...
+	subscription, err := clients.ServiceHooksClient.GetSubscription(
+		clients.Ctx,
+		servicehooks.GetSubscriptionArgs{
+			SubscriptionId: subscriptionID,
+		})
+	return subscription, err
+}
+
+// isSubscriptionDeleted checks if a service hook subscription has been deleted
+func isSubscriptionDeleted(d *schema.ResourceData, err error, subscription *servicehooks.Subscription, args *servicehooks.GetSubscriptionArgs) bool {
```
magodo: This function is not called anywhere, consider to delete it.

azuredevops/internal/service/servicehook/commons.go
```diff
@@ -0,0 +1,135 @@
...
+			Required:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.IsUUID,
+			Description:  "The ID of the project where the service hook subscription will be created",
+		},
+	}
+}
+
+// createSubscription creates a new service hook subscription in Azure DevOps
+func createSubscription(d *schema.ResourceData, clients *client.AggregatedClient, subscription *servicehooks.Subscription) (*servicehooks.Subscription, error) {
```
magodo, line 13: All these CRUD wrapper functions do nothing more than simply calling the underlying client's corresponding method, I prefer to just inlining them at where they are invoked, instead of adding this unnecessary layer.

azuredevops/internal/service/servicehook/tfs_publisher.go
```diff
@@ -0,0 +1,965 @@
...
+						Description: "Include only events for work items that contain a specific tag.",
+					},
+				},
+			},
+		},
+	}
+}
+
+// expandTfsEventConfig expands the TFS event configuration from schema
+func expandTfsEventConfig(d *schema.ResourceData) (map[string]string, string) {
```
magodo, line 564: There are quite a few boilerplates in this func body, can we refactor it a bit, e.g. introducing a helper function:

```go
// eventSelector selects the specified event config.
// Returns: event type and the raw event config
eventSelector := func(d *schema.ResourceData) (string, map[string]any) {
  for _, evt := range eventTypes {
    if l := d.Get(evt); len(l.([]any)) != 0 {
      if rawConfig, ok := l[0].(map[string]any); ok {
        return evt, rawConfig
      } else {
        // This can happen when the event config's nested fields are all optional and absent in the config
        return evt, map[string]any{}
      }
    }
  }
  return "", nil
}
```

With this, we will get a event type, which we can then `switch` and use the returned rawConfig to construct the `eventConfig` in each arm.

azuredevops/internal/service/servicehook/tfs_publisher.go
```diff
@@ -0,0 +1,965 @@
...
+		}
+	}
+
+	eventConfig["projectId"] = d.Get("project_id").(string)
+	return eventConfig, tfsResourceBlock2ApiType[eventType]
+}
+
+// flattenTfsEventConfig flattens the TFS event configuration to schema
+func flattenTfsEventConfig(subscription *servicehooks.Subscription) (string, []interface{}) {
+	eventType := tfsApiType2ResourceBlock[*subscription.EventType]
```
magodo: There needs a nil check on the input `subscription` to avoid deref on nil.

azuredevops/internal/service/servicehook/tfs_publisher.go
```diff
@@ -0,0 +1,965 @@
...
+	if isNilTfsEventConfig(*subscription.PublisherInputs) {
+		return eventType, []interface{}{nil}
```
magodo: What is the motivation here to return a `[]any{nil}`, instead of skipping this check and return `[]any{map[string]any{}}`, since each case will end up with a plan diff?

azuredevops/internal/service/servicehook/resource_servicehook_webhook_tfs.go
```diff
@@ -0,0 +1,250 @@
...
+		"http_headers": {
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "HTTP headers as key-value pairs",
+		},
+		"resource_details_to_send": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			Default:      "all",
```
magodo, line 62: Wondering if we shall make the default minimal, i.e. `none`.

azuredevops/internal/acceptancetests/resource_servicehook_webhook_tfs_test.go
```diff
@@ -0,0 +1,215 @@
+//go:build (all || resource_servicehook_webhook_tfs) && !exclude_servicehook
```
magodo: Could you please remove this line as it is not needed?

azuredevops/internal/acceptancetests/testutils/hcl.go
```diff
@@ -766,3 +766,80 @@ resource "azuredevops_servicehook_storage_queue_pipelines" "test" {
 }
 `, projectResource, accountKey, queueName, eventType)
 }
+
```
magodo: Can you move these configs from this file to its dedicated test file? This file is meant to only keep the configs that are likely shared by many other resources.

azuredevops/internal/acceptancetests/resource_servicehook_webhook_tfs_test.go
```diff
@@ -0,0 +1,215 @@
...
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: CheckServicehookWebhookTfsDestroyed,
+		Steps: []resource.TestStep{
+			{
+				Config: testutils.HclServicehookWebhookTfsResourceWithGitPushEvent(projectName, url),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(tfCheckNode, "project_id"),
+					resource.TestCheckResourceAttr(tfCheckNode, "url", url),
+					resource.TestCheckResourceAttr(tfCheckNode, "git_push.#", "1"),
```
magodo: These won't be needed if we have a import step after each test step.

Reference: https://github.com/microsoft/terraform-provider-azuredevops/blob/2dab4c90d03189c451e1a47d192b433e18973b96/azuredevops/internal/acceptancetests/resource_project_test.go#L39-L44

azuredevops/internal/acceptancetests/resource_servicehook_webhook_tfs_test.go
```diff
@@ -0,0 +1,215 @@
...
+		// indicates the subscription still exists - this should fail the test
+		if _, err := getServiceHookSubscriptionFromResource(resource); err == nil {
+			return fmt.Errorf("Unexpectedly found a service hook subscription that should be deleted")
+		}
+	}
+
+	return nil
+}
+
+func getServiceHookSubscriptionFromResource(resource *terraform.ResourceState) (interface{}, error) {
```
magodo: Can we inline this function in the call site as it is only called at one place?

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/servicehook/resource_servicehook_webhook_tfs.go
dawwestk: The Azure DevOps default is to return all, so I'd rather keep the provider in line with this behaviour

azuredevops/internal/acceptancetests/resource_servicehook_webhook_tfs_test.go
```diff
@@ -51,21 +133,25 @@ func TestAccServicehookWebhookTfs_Update(t *testing.T) {
...
 		Steps: []resource.TestStep{
 			{
-				Config: testutils.HclServicehookWebhookTfsResourceWithGitPushEvent(projectName, url1),
+				Config: HclServicehookWebhookTfsResourceWithGitPushEvent(projectName, url1),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttrSet(tfCheckNode, "project_id"),
 					resource.TestCheckResourceAttr(tfCheckNode, "url", url1),
 					resource.TestCheckResourceAttr(tfCheckNode, "git_push.#", "1"),
 				),
 			},
```
magodo, line 141: Here also wants an import step

azuredevops/internal/acceptancetests/resource_servicehook_webhook_tfs_test.go
```diff
@@ -82,20 +168,24 @@ func TestAccServicehookWebhookTfs_ChangeEventType(t *testing.T) {
...
 		Steps: []resource.TestStep{
 			{
-				Config: testutils.HclServicehookWebhookTfsResourceWithGitPushEvent(projectName, url),
+				Config: HclServicehookWebhookTfsResourceWithGitPushEvent(projectName, url),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttrSet(tfCheckNode, "project_id"),
 					resource.TestCheckResourceAttr(tfCheckNode, "git_push.#", "1"),
 				),
 			},
 			{
```
magodo, line 176: same here

magodo:
@dawwestk Thanks for updating the code! It now LGTM! Would you please update the acceptance test (currently can't compile) and pass all the tests.

dawwestk:
HI @magodo - apologies for the delay, lint and format issues are fixed and the tests for the new resources pass (i.e. `$ go test -tags "all" -v github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/service/servicehook` is successful)

magodo:
@dawwestk The unit test is still failing. Once resolved, would you please attach a screenshot/terminal paste to show the acceptance tests are passing? Thx!

dawwestk:
cheers again @magodo - here is the output of the relevant acceptance tests

```
$ TF_ACC=1 go test -timeout 120m -run ^TestAccServicehookWebhookTfs -tags "${*:-all}" -v $(go list ./... | grep acceptancetests) || fatal "Build finished in error due to failed tests"
=== RUN   TestAccServicehookWebhookTfs_basic
=== PAUSE TestAccServicehookWebhookTfs_basic
=== RUN   TestAccServicehookWebhookTfs_Update
=== PAUSE TestAccServicehookWebhookTfs_Update
=== RUN   TestAccServicehookWebhookTfs_ChangeEventType
=== PAUSE TestAccServicehookWebhookTfs_ChangeEventType
=== RUN   TestAccServicehookWebhookTfs_WithHeaders
=== PAUSE TestAccServicehookWebhookTfs_WithHeaders
=== RUN   TestAccServicehookWebhookTfs_InvalidUrl
=== PAUSE TestAccServicehookWebhookTfs_InvalidUrl
=== RUN   TestAccServicehookWebhookTfs_WithResourceDetails
=== PAUSE TestAccServicehookWebhookTfs_WithResourceDetails
=== CONT  TestAccServicehookWebhookTfs_basic
=== CONT  TestAccServicehookWebhookTfs_WithResourceDetails
=== CONT  TestAccServicehookWebhookTfs_InvalidUrl
=== CONT  TestAccServicehookWebhookTfs_ChangeEventType
=== CONT  TestAccServicehookWebhookTfs_Update
=== CONT  TestAccServicehookWebhookTfs_WithHeaders
--- PASS: TestAccServicehookWebhookTfs_InvalidUrl (2.12s)
--- PASS: TestAccServicehookWebhookTfs_WithHeaders (34.79s)
--- PASS: TestAccServicehookWebhookTfs_ChangeEventType (42.49s)
--- PASS: TestAccServicehookWebhookTfs_Update (43.02s)
--- PASS: TestAccServicehookWebhookTfs_WithResourceDetails (46.04s)
--- PASS: TestAccServicehookWebhookTfs_basic (46.61s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        46.848s
?       github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils      [no test files]
```

magodo (approved):

PR #1430, opened 2025-09-19
Adding support for work item queries

### LINKED ISSUE #1429: Feature: Add support for queries
### Community Note

### Description

It would be useful to support creating and maintaining queries. Eventually it would be nice to support adding query based widgets to a dashboard, but for now, they have utility of their own.

* azuredevops_workitemquery_folder
* azuredevops_workitemquery

```hcl
// Create a query folder
resource "azuredevops_workitemquery_folder" "query_folder" {
  project_id  = data.azuredevops_project.project.id
  name        = "Folder 1"
  parent_path = "/Shared Queries"
}

// Create a query
resource "azuredevops_workitemquery" "query" {
  project_id  = data.azuredevops_project.project.id
  name        = "Query 1"
  parent_path = azuredevops_workitemquery_folder.query_folder.path
  wiql        = "SELECT [System.Id], [System.Title] FROM WorkItems WHERE [System.AssignedTo] = @Me"
}
```

### References

* https://learn.microsoft.com/en-us/azure/devops/boards/queries/wiql-syntax?view=azure-devops
* https://learn.microsoft.com/en-us/rest/api/azure/devops/wit/queries?view=azure-devops-rest-7.2

andrwca:
These include providers for:

- A work item query folder
- A work item query

We choose to split these up into separate resources to make it simpler and easier to use. This is despite the AzDO REST API using the same APIs for both folders and queries. If we were to use a single resource for both, we'd have to expose 'isFolder', which is a little messy and prone to misconfiguration.

This implementation is fairly basic, but in testing, and in the REST API examples, it seems only the `wiql` field is read when creating and updating. A lot of the other fields seem to be read-only. In testing, the vast majority of queries can be supported, including flat, linked and recursive.

The folder resource returns it's `id` in response, which can be used in at the `parent_id` in a subsequent folder or query to create hierarchies.

## All Submissions:

Issue Number: #1429 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Other information

Closes #1429

### DIFF
--- a/azuredevops/internal/acceptancetests/resource_workitemquery_test.go
+++ b/azuredevops/internal/acceptancetests/resource_workitemquery_test.go
@@ -1,0 +1,133 @@
+//go:build (all || workitemtracking || resource_workitemquery) && (!exclude_workitemtracking || !resource_workitemquery)
+
+package acceptancetests
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
+)
+
+// Query directly under area (Shared Queries)
+func TestAccWorkItemQuery_UnderArea(t *testing.T) {
+	projectName := testutils.GenerateResourceName()
+	queryName := "tfacc-query-area"
+	wiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project"
+
+	config := testutils.HclProjectResource(projectName) + `
+resource "azuredevops_workitemquery" "query" {
+  project_id = azuredevops_project.project.id
+  name       = "` + queryName + `"
+  area       = "My Queries"
+  wiql       = "` + wiql + `"
+}
+`
+
+	res := "azuredevops_workitemquery.query"
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckProjectDestroyed,
+		Steps: []resource.TestStep{{
+			Config: config,
+			Check: resource.ComposeTestCheckFunc(
+				testutils.CheckProjectExists(projectName),
+				resource.TestCheckResourceAttr(res, "name", queryName),
+				resource.TestCheckResourceAttr(res, "wiql", wiql),
+			),
+		}},
+	})
+}
+
+// Query under a folder
+func TestAccWorkItemQuery_UnderFolder(t *testing.T) {
+	projectName := testutils.GenerateResourceName()
+	folderName := "tfacc-folder-for-query"
+	queryName := "tfacc-query-under-folder"
+	wiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project"
+
+	config := testutils.HclProjectResource(projectName) + `
+resource "azuredevops_workitemquery_folder" "folder" {
+  project_id = azuredevops_project.project.id
+  name       = "` + folderName + `"
+  area       = "My Queries"
+}
+
+resource "azuredevops_workitemquery" "query" {
+  project_id = azuredevops_project.project.id
+  name       = "` + queryName + `"
+  parent_id  = azuredevops_workitemquery_folder.folder.id
+  wiql       = "` + wiql + `"
+}
+`
+
+	folderRes := "azuredevops_workitemquery_folder.folder"
+	queryRes := "azuredevops_workitemquery.query"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckProjectDestroyed,
+		Steps: []resource.TestStep{{
+			Config: config,
+			Check: resource.ComposeTestCheckFunc(
+				testutils.CheckProjectExists(projectName),
+				resource.TestCheckResourceAttr(folderRes, "name", folderName),
+				resource.TestCheckResourceAttr(queryRes, "name", queryName),
+				resource.TestCheckResourceAttr(queryRes, "wiql", wiql),
+			),
+		}},
+	})
+}
+// Update existing query (name + WIQL)
+func TestAccWorkItemQuery_Update(t *testing.T) {
+	projectName := testutils.GenerateResourceName()
+	initialName := "tfacc-query-update-initial"
+	updatedName := "tfacc-query-update-updated"
+	initialWiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project"
+	updatedWiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project ORDER BY [System.Id] DESC"
+
+	configCreate := testutils.HclProjectResource(projectName) + `
+resource "azuredevops_workitemquery" "query" {
+  project_id = azuredevops_project.project.id
+  name       = "` + initialName + `"
+  area       = "My Queries"
+  wiql       = "` + initialWiql + `"
+}
+`
+
+	configUpdate := testutils.HclProjectResource(projectName) + `
+resource "azuredevops_workitemquery" "query" {
+  project_id = azuredevops_project.project.id
+  name       = "` + updatedName + `"
+  area       = "My Queries"
+  wiql       = "` + updatedWiql + `"
+}
+`
+
+	res := "azuredevops_workitemquery.query"
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckProjectDestroyed,
+		Steps: []resource.TestStep{
+			{
+				Config: configCreate,
+				Check: resource.ComposeTestCheckFunc(
+					testutils.CheckProjectExists(projectName),
+					resource.TestCheckResourceAttr(res, "name", initialName),
+					resource.TestCheckResourceAttr(res, "wiql", initialWiql),
+				),
+			},
+			{
+				Config: configUpdate,
+				Check: resource.ComposeTestCheckFunc(
+					testutils.CheckProjectExists(projectName),
+					resource.TestCheckResourceAttr(res, "name", updatedName),
+					resource.TestCheckResourceAttr(res, "wiql", updatedWiql),
+				),
+			},
+		},
+	})
+}
--- a/azuredevops/internal/acceptancetests/resource_workitemquery_test.go
+++ b/azuredevops/internal/acceptancetests/resource_workitemquery_test.go
@@ -1,8 +1,7 @@
-//go:build (all || workitemtracking || resource_workitemquery) && (!exclude_workitemtracking || !resource_workitemquery)
-
 package acceptancetests
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
@@ -15,14 +14,7 @@ func TestAccWorkItemQuery_UnderArea(t *testing.T) {
 	queryName := "tfacc-query-area"
 	wiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project"
 
-	config := testutils.HclProjectResource(projectName) + `
-resource "azuredevops_workitemquery" "query" {
-  project_id = azuredevops_project.project.id
-  name       = "` + queryName + `"
-  area       = "My Queries"
-  wiql       = "` + wiql + `"
-}
-`
+	config := hclWorkItemQueryResource(projectName, queryName, wiql)
 
 	res := "azuredevops_workitemquery.query"
 	resource.ParallelTest(t, resource.TestCase{
@@ -47,20 +39,7 @@ func TestAccWorkItemQuery_UnderFolder(t *testing.T) {
 	queryName := "tfacc-query-under-folder"
 	wiql := "SELECT [System.Id] FROM WorkItems WHERE [System.TeamProject] = @project"
 
-	config := testutils.HclProjectResource(projectName) + `
-resource "azuredevops_workitemquery_folder" "folder" {
-  project_id = azuredevops_project.project.id
-  name       = "` + folderName + `"
-  area       = "My Queries"
-}
-
-resource "azuredevops_workitemquery" "query" {
-  project_id = azuredevops_project.project.id
-  name       = "` + queryName + `"
-  parent_id  = azuredevops_workitemquery_folder.folder.id
-  wiql       = "` + wiql + `"
-}
-`
+	config := hclWorkItemQueryUnderFolderResource(projectName, folderName, queryName, wiql)
 
 	folderRes := "azuredevops_workitemquery_folder.folder"
 	queryRes := "azuredevops_workitemquery.query"
@@ -80,6 +59,7 @@ resource "azuredevops_workitemquery" "query" {
 		}},
 	})
 }
+
 // Update existing query (name + WIQL)
 func TestAccWorkItemQuery_Update(t *testing.T) {
 	projectName := testutils.GenerateResourceName()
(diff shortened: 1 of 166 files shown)


andrwca:
Question: should we support `data` for `azuredevops_workitemquery`? Might be useful to get the `wiql` or `path` back from an existing query?

azuredevops/internal/service/queries/resource_workitemquery.go
```diff
@@ -0,0 +1,185 @@
+package queries
```
magodo: These two files shall go to the `workitemtracking` package instead.

azuredevops/internal/service/queries/resource_workitemquery_folder.go
```diff
@@ -0,0 +1,132 @@
...
+			},
+
+			"project_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"parent_path": {
```
magodo: Instead of using `path`, we shall insist on using `parent_id`, as the other resources do in this provider.

azuredevops/internal/service/queries/resource_workitemquery_folder.go
```diff
@@ -0,0 +1,132 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringLenBetween(1, 256),
+			},
+
+			// Add 'path' attribute to store the full path of the query folder. This is read-only and computed.
+			"path": {
```
magodo: Based on above, this is not needed then.

azuredevops/internal/service/workitemtracking/resource_workitemquery_folder.go
```diff
@@ -0,0 +1,132 @@
...
+			IsFolder: converter.Bool(true),
+		},
+	}
+
+	resp, err := clients.WorkItemTrackingClient.CreateQuery(clients.Ctx, params)
+	if err != nil {
+		return diag.Errorf(" Creating query folder. Error: %s", err)
+	}
+
+	d.SetId(resp.Id.String())
```
magodo, line 93: We shall nil check `resp.Id` here.

azuredevops/internal/service/queries/resource_workitemquery.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+
+			"project_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+			},
+
+			"parent_path": {
```
magodo: Please consider to use `parent_id` instead.

azuredevops/internal/service/queries/resource_workitemquery.go
```diff
@@ -0,0 +1,185 @@
...
+			},
+
+			"wiql": {
+				Type:     schema.TypeString,
+				Required: true,
+				// The value of 32000 matches the restrictions in Azure DevOps.
+				ValidateFunc: validation.StringLenBetween(1, 32000),
+			},
+
+			"path": {
```
magodo: This is less useful.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,185 @@
...
+			IsFolder: converter.Bool(false),
+		},
+	}
+
+	resp, err := clients.WorkItemTrackingClient.CreateQuery(clients.Ctx, params)
+	if err != nil {
+		return diag.Errorf(" Creating query. Error: %s", err)
+	}
+
+	d.SetId(resp.Id.String())
```
magodo, line 105: Nil check `resp.Id`.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,185 @@
...
+		if resp.Path != nil {
+			d.Set("path", resp.Path)
+		}
+		if resp.Name != nil {
+			d.Set("name", resp.Name)
+		}
+		if resp.Wiql != nil {
+			d.Set("wiql", resp.Wiql)
+		}
+	}
```
magodo, line 137: Shall we also check that this is a non-folder here?

azuredevops/internal/service/workitemtracking/resource_workitemquery_folder.go
```diff
@@ -0,0 +1,132 @@
...
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
```
magodo, line 31: Arguably, the `name` can be updated?

magodo (changes requested):
Thanks for this PR! I've taken a look through and left some comments inline. Also, please add the corresponding tests and documents.

Regarding the data source, it is nice to have, but is fine to leave it as for now.

andrwca:
> Thanks for this PR! I've taken a look through and left some comments inline. Also, please add the corresponding tests and documents.
> 
> Regarding the data source, it is nice to have, but is fine to leave it as for now.

@magodo I'm just trying to work out how these changes would look. 

If we changed to using `parent_id` instead of `parent_path` it does make things a little more complicated, based on how the REST API works and how there are two immutable root folders (areas) under which everything else has to be placed.

One way would be for work item query folder to support either `parent_id`, or if it's a folder under one of the root folders, `area` which can be only `My Queries` or `Shared Queries` 

Or we could support both `parent_id` and `parent_path` for both use cases.

The former option would look like this:

```tf
// Create a query folder under My Queries
resource "azuredevops_workitemquery_folder" "query_folder1" {
  project_id = data.azuredevops_project.project.id
  area       = "My Queries"
  name       = "Folder 1"
}

// Create another query folder under the first folder
resource "azuredevops_workitemquery_folder" "query_folder2" {
  project_id = data.azuredevops_project.project.id
  parent_id  = data.azuredevops_workitemquery_folder.query_folder1.id
  name       = "Folder 2"
}

// Create a query
resource "azuredevops_workitemquery" "query" {
  project_id = data.azuredevops_project.project.id
  parent_id  = azuredevops_workitemquery_folder.query_folder2.id
  name       = "Query 1"
  wiql       = "SELECT [System.Id], [System.Title] FROM WorkItems WHERE [System.AssignedTo] = @Me"
}

data "azuredevops_group" "example-readers" {
  project_id = data.azuredevops_project.project.id
  name       = "Readers"
}

// Add permissions to the query
resource "azuredevops_workitemquery_permissions" "query_permissions" {
  project_id = data.azuredevops_project.project.id
  path       = "${azuredevops_workitemquery.query_folder1.name}/${azuredevops_workitemquery.query_folder2.name}/${azuredevops_workitemquery.query.name}"
  principal  = data.azuredevops_group.example-readers.id
  permissions = {
    "View"   = "Allow"
    "Edit"   = "Allow"
    "Manage" = "Allow"
    "Delete" = "Deny"
  }
}
```

Any thoughts?

andrwca:
Just thinking, the work item query resource will need to support `area` too for queries that sit directly under `My Queries` or `Shared Queries`.

magodo:
@andrwca Is there a way to query the id for the two root folders using a data source? Based on your above example, it makes me feel the `path` attribute is still useful to be referenced in `azuredevops_workitemquery_permissions`.

andrwca:
@magodo Yes, you'd be able to.

Only thing is, for the simplest use case (a single folder/query under `Shared Queries` or `My Queries`), you need two resources instead of one, and in all cases, you'd need to use a data source to get either the `Shared Queries` or `My Queries` folder. And if you omitted `parent_id` and used a name other than `Shared Queries` or `My Queries`, it'd be invalid (they are the only two folders that can exist at the root), hence a suggestion for the `area` property which limits the input string to valid options. The `area` field is essentially a convenience for the need to use a data source.

So without area

```tf
data "azuredevops_workitemquery_folder" "query_folder_my_queries" {
  project_id = data.azuredevops_project.project.id
  name       = "My Queries"
}
```

```tf
resource "azuredevops_workitemquery_folder" "query_folder1" {
  project_id = data.azuredevops_project.project.id
  parent_id = data.azuredevops_workitemquery_folder.query_folder_my_queries
  name       = "Folder 1"
}
```

With area

```tf
resource "azuredevops_workitemquery_folder" "query_folder1" {
  project_id = data.azuredevops_project.project.id
  area       = "My Queries"
  name       = "Folder 1"
}
```

azuredevops/internal/service/workitemtracking/resource_workitemquery_folder.go
andrwca: Not in the folder resource, no. We'll need to destroy the folder, cascading down to any resources underneath and recreate them afterwards. The query resource allows `name` updates though.

magodo:
All right, let's introduce the `area` then. Please ensure to add some basic validation for it though, e.g. `area` and `parent_id` are in conflict.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+			},
+
+			// The ID of the parent folder.
+			// It should not be specified if 'area' is specified.
+			"parent_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ValidateFunc:  validation.IsUUID,
+				ConflictsWith: []string{"area"},
```
magodo: We shall use `ExactlyOneOf` instead to indicate that one of them is required.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+			},
+
+			// If specified, the area should be one of either 'Shared Queries' or 'My Queries'.
+			// It should not be specified if 'parent_id' is specified.
+			"area": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ForceNew:      true,
+				ValidateFunc:  validation.StringInSlice([]string{"Shared Queries", "My Queries"}, false),
+				ConflictsWith: []string{"parent_id"},
```
magodo: We shall use `ExactlyOneOf` instead to indicate that one of them is required.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+
+func resourceQueryCreate(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	clients := m.(*client.AggregatedClient)
+
+	projectID := d.Get("project_id").(string)
+
+	// A parent ID and area cannot be set together.
+	if d.Get("area").(string) != "" && d.Get("parent_id").(string) != "" {
+		return diag.Errorf("Only one of 'area' or 'parent_id' should be set.")
+	}
```
magodo: With the schema constraint, these are not needed then.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+			Wiql:     converter.String(d.Get("wiql").(string)),
+			IsFolder: converter.Bool(false),
+		},
+	}
+
+	resp, err := clients.WorkItemTrackingClient.CreateQuery(clients.Ctx, params)
+	if err != nil {
+		return diag.Errorf(" Creating query. Error: %s", err)
+	}
+
```
magodo, line 96: Can we also nil check the `resp`?

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		return diag.Errorf(" Getting query with id: %s. Error: %+v", id, err)
+	}
+
+	// Check that the query is not a folder
+	if resp.IsFolder != nil && *resp.IsFolder {
+		return diag.Errorf(" The query with id: %s is a folder. Expected a query.", id)
+	}
+
+	if resp != nil {
```
magodo, line 133: Can we move this check above the L141?

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+	params := workitemtracking.GetQueryArgs{
+		Project: converter.String(d.Get("project_id").(string)),
+		Query:   &id,
+	}
+
+	existing, err := clients.WorkItemTrackingClient.GetQuery(clients.Ctx, params)
+	if err != nil {
+		if utils.ResponseWasNotFound(err) {
+			d.SetId("")
+			return nil
```
magodo: During Update, if the read returns 404, we don't remove it from state, but just throw error.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
```diff
@@ -0,0 +1,216 @@
...
+
+	_, err = clients.WorkItemTrackingClient.UpdateQuery(clients.Ctx, updateArgs)
+	if err != nil {
+		return diag.Errorf(" Updating query with ID: %s. Error detail: %+v", id, err)
+	}
+
+	readDiags := resourceQueryRead(clients.Ctx, d, m)
+	diags = append(diags, readDiags...)
+
+	return diags
```
magodo:
```suggestion
return resourceQueryRead(clients.Ctx, d, m)
```

azuredevops/internal/acceptancetests/resource_workitemquery_test.go
```diff
@@ -0,0 +1,128 @@
...
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckProjectDestroyed,
+		Steps: []resource.TestStep{{
+			Config: config,
+			Check: resource.ComposeTestCheckFunc(
+				testutils.CheckProjectExists(projectName),
+				resource.TestCheckResourceAttr(res, "name", queryName),
+				resource.TestCheckResourceAttr(res, "wiql", wiql),
+			),
+		}},
```
magodo, line 31: Can we add an import step after each "regular" test step to validate the import functionality also works?

(This applies to every test case)

Refer to the *azuredevops/internal/acceptancetests/resource_project_test.go* for an example.

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
andrwca: Good catch, yup.

azuredevops/internal/service/workitemtracking/resource_workitemquery.go
andrwca: Fixed!

azuredevops/internal/acceptancetests/resource_workitemquery_test.go
andrwca: @magodo I spent a bit of time looking into supporting `import` properly, and it's actually a little more complicated than I have time for, due to the options of using `parent_id` or `area`. I think it could be quite easy to implement it in a broken way if it's rushed, and I unfortunately don't have the time it needs, so I'd suggest adding this to the backlog if it's needed. 

I've updated the docs anyway to say that this resource doesn't support import, and if you try to use `terraform import` you'll get a clear message explaining that it's not supported.

I've also removed reading the `wiql` property on read as it was never returned by the client as the request wasn't ever expanded. If you do expand the request and update the `wiql` property, the problem then arises that the `wiql` property needs to be normalized to make sense. This is because AzDO will format the WIQL and return a differently formatted string in response to further REST API calls.

Again, this is an area that might be easy to get wrong if rushed, so I'd suggest not doing it in this PR at this time and accepting the limitation.

I have sorted all the other comments though. :)

magodo (approved):

magodo:
Tests are passing:

```shell
$ TF_ACC=1 go test -v -run='TestAccWorkItemQueryFolder_|TestAccWorkItemQuery_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItemQueryFolder_UnderArea
=== PAUSE TestAccWorkItemQueryFolder_UnderArea
=== RUN   TestAccWorkItemQueryFolder_UnderFolder
=== PAUSE TestAccWorkItemQueryFolder_UnderFolder
=== RUN   TestAccWorkItemQuery_UnderArea
=== PAUSE TestAccWorkItemQuery_UnderArea
=== RUN   TestAccWorkItemQuery_UnderFolder
=== PAUSE TestAccWorkItemQuery_UnderFolder
=== RUN   TestAccWorkItemQuery_Update
=== PAUSE TestAccWorkItemQuery_Update
=== CONT  TestAccWorkItemQueryFolder_UnderArea
=== CONT  TestAccWorkItemQuery_UnderFolder
=== CONT  TestAccWorkItemQuery_Update
=== CONT  TestAccWorkItemQuery_UnderArea
=== CONT  TestAccWorkItemQueryFolder_UnderFolder
--- PASS: TestAccWorkItemQueryFolder_UnderArea (27.21s)
--- PASS: TestAccWorkItemQuery_UnderArea (27.40s)
--- PASS: TestAccWorkItemQueryFolder_UnderFolder (28.11s)
--- PASS: TestAccWorkItemQuery_UnderFolder (28.24s)
--- PASS: TestAccWorkItemQuery_Update (31.41s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        31.430s
```

Issue #1431, opened 2025-09-21
Feature: azuredevops_group skip_destroy


guderkar:
We have permissions in Azure DevOps scoped to specific project.

What we need is to fetch/sync entra group by object id to get ADO descriptor.

We are using `azuredevops_group` resource since there is not data source for such a purpose (https://github.com/microsoft/terraform-provider-azuredevops/issues/520).

The creation of the resource works for us, also as users we are able to fetch entra groups and add them as members to ADO groups in web UI.

```
resource "azuredevops_group" "example" {
  origin_id   = "__entra_group_object_id__"
}
```

The problem comes with the deletion of the resource, where we get permissions denied.
Which is understandable since we have such scoped permissions.
```
Error: Waiting for group delete. Access Denied: XXX-XXX-XXX-XXX-XXX needs the following permission(s) in the Identity security namespace to perform this action: Delete identity information. Additional details: security token: XXX-XXX-XXX-XXX-XXX\XXX-XXX-XXX-XXX-XXX, permission bits required: 4.
```

**Proposal**
Add meta argument which will make sure the API for deletion is not called, just remove the resource from the state.

```
resource "azuredevops_group" "example" {
  origin_id   = "__entra_group_object_id__"
  skip_destroy = true
}
```

magodo:
Maybe this is what you want: https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/data-sources/descriptor.

guderkar:
I already checked this one, but this is same problem as getting the descriptor, i.e. storage_key is internal ID of azure devops which I can get only via syncing through `azuredevops_group` resource first.

magodo:
As is discussed in #1441, the feature should be supported at the terraform core (https://github.com/hashicorp/terraform/issues/3874) or opentofu (https://github.com/opentofu/opentofu/pull/3409/files). Closing this issue as we don't plan to support it at the provider level.

guderkar:
FYI I found another (little bit dirty) way to work around the issue using HTTP data source to get the ADO descriptor.

```
data "http" "devops_groups" {
  for_each = { for group in var.config.groups : group.group_name => group }

  url = "https://vssps.dev.azure.com/${var.config.ado_organization_name}/_apis/graph/groups?api-version=7.1-preview.1"

  method = "POST"

  request_headers = {
    Content-Type  = "application/json"
    Authorization = "Bearer ${var.ado_token}"
  }

[...]

PR #1441, opened 2025-10-16
Added optional skip_destroy meta argument for azuredevops_group resource

guderkar:

This pull request adds `skip_destroy` option for resource `azuredevops_group`.

This option is useful for cases when you need to sync Entra ID backed group to Azure DevOps with identity which might not have sufficient organization level privileges to delete the group afterwards.

Issue: https://github.com/microsoft/terraform-provider-azuredevops/issues/1431

- [ ] Yes
- [*] No

- [ ] Yes
- [*] No

### DIFF
--- a/azuredevops/internal/service/graph/resource_group.go
+++ b/azuredevops/internal/service/graph/resource_group.go
@@ -83,6 +83,13 @@ func ResourceGroup() *schema.Resource {
 				Set:        schema.HashString,
 			},
 
+			"skip_destroy": {
+				Type:          schema.TypeBool,
+				Optional:      true,
+				Default:       false,
+				ConflictsWith: []string{"mail", "display_name", "scope"},
+			},
+
 			"url": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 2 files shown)


guderkar:

magodo:
Does https://developer.hashicorp.com/terraform/tutorials/state/resource-lifecycle#prevent-resource-deletion work in this case?

guderkar:
If you are asking if prevent_destroy will take precedence, then yes it will.

Although those two options together do not make much sense.

image

magodo:
Sorry, there is another (top) feature that is requested in `terraform`: https://github.com/hashicorp/terraform/issues/2253. I'd be refrained to introduce a per resource flag for this feature. Alternatively, the `opentofu` has the [support to exclude a resource when apply](https://opentofu.org/blog/opentofu-1-9-0-beta1/#the--exclude-flag), if that is an option to you.

guderkar:
Yes true I would also rather see such a feature implemented in terraform extending the lifecycle functionality. But something tells me it may take years.

guderkar:
Lol opentofu did it 4 days ago https://github.com/opentofu/opentofu/pull/3409/files.

Terraform seems to track this issue here https://github.com/hashicorp/terraform/issues/3874, the title is little bit unfortunate and since the issue is here already for 10 years I doubt they will ever pull it off.

magodo:
@guderkar Right, they released this feature today. May I close this issue since there is no plan to skip a certain resource type at the provider level? Thank you for your understanding.

guderkar:
Yes the issue can be closed thx.

Issue #1446, opened 2025-11-11
Feature Request: Add support for Inherited Processes and Custom Work Item Types (Azure Boards)


Fresa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The Azure DevOps Terraform provider currently lacks resources for managing inherited processes and custom work item types (WITs) in Azure Boards.

While Azure DevOps REST APIs expose management capabilities for processes (e.g., /processes, /processes/{processId}/workItemTypes, /fields, /states, etc.), there are no corresponding Terraform resources to declaratively manage them.

This makes it impossible to treat process configuration as code using Terraform — unlike projects, repos, service connections, and pipelines that are already supported.

### New or Affected Resource(s)

* azuredevops_process
* azuredevops_work_item_type

There might very well be more types here that are needed.

### Potential Terraform Configuration

```hcl
resource "azuredevops_process" "inherited" {
  name         = "Acme - Inherited"
  inherits_from = "Agile"
  description  = "Custom process for Acme projects"
}

resource "azuredevops_work_item_type" "ticket" {
  process_id   = azuredevops_process.inherited.id
  name         = "Ticket"
  description  = "Customer support ticket"
  fields = [
    {
      reference_name = "Custom.CustomerImpact"
      name           = "Customer Impact"
      type           = "string"
    }
  ]
  states = [
    { name = "New" },
    { name = "Triage" },
    { name = "Resolved" },
  ]
}
```

### References

Azure DevOps REST API: [Work Item Types](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/work-item-types?view=azure-devops-rest-7.1)

Azure DevOps REST API: [Processes](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/processes?view=azure-devops-rest-7.1)

Existing provider docs: [https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs)

dennismdejong:
[...]

Issue #438, opened 2021-08-21
Support for process template


roryprimrose:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

We have a custom process template that includes several new fields and work items. For fields we have defined pick lists and for work items we have custom layouts, states and rules. We also need to replicate this template over several organisations. Terraform would be perfect to make this consistent. 

### New or Affected Resource(s)

* azuredevops_field
* azuredevops_process
* azuredevops_picklist
* azuredevops_process_workitem
* azuredevops_process_state
* azuredevops_process_rule

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

* #0000

Fresa:
FYI
https://github.com/microsoft/terraform-provider-azuredevops/issues/1446
https://github.com/microsoft/terraform-provider-azuredevops/issues/1488
https://github.com/microsoft/terraform-provider-azuredevops/issues/839

PR #1454, opened 2025-11-24
Add support for workitemtrackingprocess processes

Fresa:
## All Submissions:

Introduces the resource `azuredevops_workitemtrackingprocess_process` and corresponding data sources `azuredevops_workitemtrackingprocess_process` and `azuredevops_workitemtrackingprocess_processes`.

- [x] Yes
- [ ] No

Added dependency on `github.com/microsoft/azure-devops-go-api/azuredevops/v7/workitemtrackingprocess`.
Direct dependency on `github.com/google/go-cmp v0.6.0`.

- [ ] Yes
- [x] No

N/A

## Other information
There are a couple of unrelated unit tests failing, but they are failing on main to.


Fresa:

dennismdejong:
Is also covering #438

azuredevops/internal/website-scaffold/main.go
```diff
@@ -205,11 +205,6 @@ func (gen documentationGenerator) argumentsBlock() string {
 		for _, fieldName := range gen.sortFields(input) {
```
magodo, line 205: Any reason to change this file?

azuredevops/internal/utils/testhelper/schema.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"testing"
+
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Validates that the resource data conforms to the schema
+func ValidateResourceData(t *testing.T, d *schema.ResourceData, r *schema.Resource) {
```
magodo: This is not necessary as the acceptance test will cover it.

website/docs/r/workitemtrackingprocess_process.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+* `parent_process_type_id` - (Required)  ID of the parent process. Changing this forces a new process to be created.
+
+---
+
+* `description` - (Optional)  Description of the process.
+
+* `expand` - (Optional)  Specifies the expand option when getting the process.
+
+* `is_default` - (Optional)  Is the process default?
```
magodo: Can we mention the default value?

website/docs/r/workitemtrackingprocess_process.html.markdown
```diff
@@ -0,0 +1,82 @@
...
+
+* `is_enabled` - (Optional)  Is the process enabled?
```
magodo: Can we mention the default value?

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
```diff
@@ -0,0 +1,243 @@
...
+			},
+			"expand": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "none",
+				ValidateFunc: validation.StringInSlice([]string{"none", "projects"}, false),
+				Description:  "Specifies the expand option when getting the process.",
+			},
+			"projects": {
+				Type: schema.TypeSet,
```
magodo: This needs to be TypeList to be computed.

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
```diff
@@ -0,0 +1,243 @@
...
+				Optional:    true,
+				Default:     true,
+				Description: "Is the process enabled?",
+			},
+			"customization_type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Indicates the type of customization on this process. System Process is default process. Inherited Process is modified process that was System process before.",
+			},
+			"expand": {
```
magodo: This is a behavior like attribute instead of a concrete attribute of this resource, can we either always expand it, or we can just leave the `projects` away as it might be not interesting to the users?

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
```diff
@@ -0,0 +1,130 @@
...
+			},
+		},
+	})
+}
+
+func process(name string) string {
+	return fmt.Sprintf(`
+resource "azuredevops_workitemtrackingprocess_process" "test" {
+  name                   = "%s"
+  parent_process_type_id = "adcc42ab-9882-485e-a3ed-7678f01f66bc" // Wellknown system process
```
magodo: Will this change for different organization/ADO isntance? If no, then can we just hardcode it in the code, or is there any data source that we can read it from dynamically instead of hardcoding it here in the test?

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
```diff
@@ -0,0 +1,243 @@
...
+
+	isDefault := d.Get("is_default").(bool)
+	isEnabled := d.Get("is_enabled").(bool)
+	if *processInfo.IsDefault != isDefault ||
+		*processInfo.IsEnabled != isEnabled {
+		err := updateResourceProcess(ctx, d, m)
+		if err != nil {
+			return err
+		}
+		// Circumvent eventual consistency of the read api as there is no proper way of dealing with this upstream
```
magodo: Does this mean the read right after update will not reflect the status? If that is the case, is that also an issue for the Update function, where you didn't set these attribute, but delegate to the Read function instead?

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/website-scaffold/main.go
Fresa: I added support to source description for documentation from the schema

```
if field.Description != "" {
	return fmt.Sprintf("%s %s", generateBlockHint(), field.Description)
}
```

azuredevops/internal/utils/testhelper/schema.go
Fresa: I added this after I found myself forgetting to define required fields when setting up input for unit tests several times, it helped me write unit tests a bit faster

website/docs/r/workitemtrackingprocess_process.html.markdown
Fresa: yepp, will fix! I will go through all optional fields with default values and adjust.

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: What does that mean? I chose TypeSet as I interpret the list to be unordered coming from the downstream API

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: yeah I like that, it would reduce complexity, and I can't really see a use case either. It's also possible to fetch unmanaged projects via `data.azuredevops_workitemtrackingprocess_process`. I'll remove it!

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
Fresa: I haven't found any documentation stating this as a fact, but when googling the UUID, there are a lot of results hinting that it might be hardcoded. Even the [example for the GET api endpoint for processes hint this](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/processes/get?view=azure-devops-rest-7.1&tabs=HTTP#:~:text=api%2Dversion%3D7.1-,Sample%20response,-Status%20code%3A).

All processes (including system defined) can be read via `data.azuredevops_workitemtrackingprocess_processes", but one would need to identify the process wanted somehow (id, name etc).

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: I think this is a general problem affecting most entities within the DevOps API. I couldn't find any mechanism for detecting inconsistency in the API (like ETag), this was the simplest solution I was able to stir up.

I detected this problem by running one of the acceptance test many times.

website/docs/r/workitemtrackingprocess_process.html.markdown
Fresa: Done. I also added support to the document generator to include default values defined in the schema.

website/docs/r/workitemtrackingprocess_process.html.markdown
Fresa: It's mentioned in the description that the server will assign a unique reference name if non is specified

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: I rewrote both create and update to use the response from the respective operation instead of calling read, that seems to work much better, and it's simpler

azuredevops/internal/website-scaffold/main.go
magodo: Since this will impact all generated document, and the description field is not consistently added among all resources. Can we put this change aside first?

azuredevops/internal/utils/testhelper/schema.go
magodo: That said, the mistake will be caught by acctest anyway. It helps the development, while it is not necessary to be part of the final PR.

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
magodo: Normally we shall avoid using a hardcoded value even though it is a wellknown id. Would you mind to explore if there is any way to reference this via the data source?

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func flattenProcess(d *schema.ResourceData, process *workitemtrackingprocess.ProcessInfo) diag.Diagnostics {
+	d.Set("name", process.Name)
+	d.Set("description", process.Description)
+	d.Set("parent_process_type_id", process.ParentProcessTypeId.String())
+	d.Set("reference_name", process.ReferenceName)
+	d.Set("is_default", process.IsDefault)
+	d.Set("is_enabled", process.IsEnabled)
+	d.Set("customization_type", string(*process.CustomizationType))
```
magodo: These require nil check since they are pointers.

magodo (changes requested):

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
magodo: I actually tried your branch with changing the Create/Update return line to return the calling of the `readResourceProcess()`. It turns out everything works well. I'm wondering what is the issue that you've met.

Fresa: If you tried the old changes (that I've refactored since), then you might need to run it a couple of times to catch it

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
Fresa: I've spent some more time researching, but one has to assume something to identify the system process to inherit from. If not the id, then the name for example:

```
# Get all processes to find the Agile system process
data "azuredevops_workitemtrackingprocess_processes" "all" {
}

# Extract the Agile process ID
locals {
  agile_process_id = [
    for process in data.azuredevops_workitemtrackingprocess_processes.all.processes :
    process.id if process.name == "Agile" && process.customization_type == "System"
  ][0]
}

# Create a custom process based on Agile
resource "azuredevops_workitemtrackingprocess_process" "custom" {
  name                   = "My Custom Agile Process"
  parent_process_type_id = local.agile_process_id
  description            = "Custom process based on Agile"
}
```

But I think it's more likely that the name of a system process might change than the id.

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: `d.Set` handles nil (it will use the zero value), but I explicitly didn't add any nil checks here because non of those properties are expected to be nil. Unfortunately the documentation does not explicitly declare required properties of anything.

I looked at other resources as example for conventions in this repo, and this seemed to be the preferred way, but I can change it if you like? There is also the matter of `Set` returning an error, should we also deal with that for primitives?

azuredevops/internal/website-scaffold/main.go
Fresa: fair enough, I've reverted all changes

azuredevops/internal/utils/testhelper/schema.go
Fresa: I've reverted it!

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
magodo: Please add the nil check as the API behavior is undetermined. For error checking on `d.Set`, it is only needed when setting complicated (i.e. not primitive) types.

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
magodo: I've tried a couple times, all good. Would you please share the detail of the error you hit?

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
magodo: I'm not sure, then let's use the id instead, with a comment mentioning it is sourced from the official example (with a link).

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: Here's a couple of examples using [this commit](https://github.com/Fresa/terraform-provider-azuredevops/commit/684f23b446da6c14484b7f2de3261e1be155a159).

```
Running tool: /usr/local/go/bin/go test -timeout 30m -run ^TestAccWorkitemtrackingprocessProcess_CreateDisabled$ github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests -v -tags=all -args -test.v

=== RUN   TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== PAUSE TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== CONT  TestAccWorkitemtrackingprocessProcess_CreateDisabled
    /home/fredrik/go/src/github.com/Fresa/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testing_new.go:85: Error running post-test destroy, there may be dangling resources: process with ID c6c2b4bf-f7d5-456f-a69e-14ed1c404e98 should not exist
--- FAIL: TestAccWorkitemtrackingprocessProcess_CreateDisabled (4.40s)
FAIL
FAIL	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	4.418s
```
Reading after deleting fails intermittently due to eventual consistency.

```
Running tool: /usr/local/go/bin/go test -timeout 30m -run ^TestAccWorkitemtrackingprocessProcess_CreateAndUpdate$ github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests -v -tags=all -args -test.v

=== RUN   TestAccWorkitemtrackingprocessProcess_CreateAndUpdate
=== PAUSE TestAccWorkitemtrackingprocessProcess_CreateAndUpdate
=== CONT  TestAccWorkitemtrackingprocessProcess_CreateAndUpdate
    /home/fredrik/go/src/github.com/Fresa/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go:72: Step 3/4 error: Check failed: Check 1/1 error: azuredevops_workitemtrackingprocess_process.test: Attribute 'is_enabled' expected "false", got "true"
    /home/fredrik/go/src/github.com/Fresa/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testing_new.go:85: Error running post-test destroy, there may be dangling resources: process with ID 3e6a6326-5267-44b0-ae64-e5a2d9cd1fb7 should not exist
--- FAIL: TestAccWorkitemtrackingprocessProcess_CreateAndUpdate (4.79s)
FAIL
FAIL	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	4.810s
```
Same as above + similar problem with reading after update.

Comment out these statements:
```
// d.Set("is_default", isDefault)
// d.Set("is_enabled", isEnabled)
// return nil
```

```
Running tool: /usr/local/go/bin/go test -timeout 30m -run ^TestAccWorkitemtrackingprocessProcess_CreateDisabled$ github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests -v -tags=all -args -test.v

=== RUN   TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== PAUSE TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== CONT  TestAccWorkitemtrackingprocessProcess_CreateDisabled
    /home/fredrik/go/src/github.com/Fresa/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go:46: Step 1/2 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

          # azuredevops_workitemtrackingprocess_process.test will be updated in-place
          ~ resource "azuredevops_workitemtrackingprocess_process" "test" {
                id                     = "a89b2ea2-9f50-439e-954f-967194e8bfd9"
              ~ is_enabled             = true -> false
                name                   = "test-acc-dwx39hiz8e"
                # (5 unchanged attributes hidden)
            }

        Plan: 0 to add, 1 to change, 0 to destroy.
--- FAIL: TestAccWorkitemtrackingprocessProcess_CreateDisabled (3.17s)
FAIL
FAIL	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	3.186s
```
I think it's the import step that causes this when reading (right after update) during refresh. This is actually still an issue, I can reproduce it on the latest commit, however it seems to happen less often, and it's only really going to be an issue in the acceptance tests.

This test may also fail similar as the create+update one above:
```
Running tool: /usr/local/go/bin/go test -timeout 30m -run ^TestAccWorkitemtrackingprocessProcess_CreateDisabled$ github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests -v -tags=all -args -test.v

=== RUN   TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== PAUSE TestAccWorkitemtrackingprocessProcess_CreateDisabled
=== CONT  TestAccWorkitemtrackingprocessProcess_CreateDisabled
    /home/fredrik/go/src/github.com/Fresa/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go:46: Step 1/2 error: Check failed: Check 2/2 error: azuredevops_workitemtrackingprocess_process.test: Attribute 'is_enabled' expected "false", got "true"
--- FAIL: TestAccWorkitemtrackingprocessProcess_CreateDisabled (2.25s)
FAIL
FAIL	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	2.270s
```

If I add `time.Sleep(1 * time.Second)` before reading the resource, I can't reproduce the above anymore.

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: Done!

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
Fresa: I've added a link to the example.

I'm wondering if we should perhaps rename `parent_process_type_id` to `inherited_system_process_type_id` or something similar instead, as [system processes are the only ones that can be inherited](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/inheritance-process-model?view=azure-devops&tabs=agile-process#system-and-inherited-processes). What do you think?

Perhaps even rename this resource to `azuredevops_workitemtrackingprocess_inherited_process`? 

The data sources are however able to return both system and inherited processes.

magodo: I'm not sure, normally I'll follow the API name in case it gets extended in the future to support other process type?

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
magodo: You are right, this is an eventual consistency issue. Let's keep setting these fields here as you did then.

azuredevops/internal/service/workitemtrackingprocess/data_process.go
```diff
@@ -0,0 +1,154 @@
...
+				Computed:    true,
+				Description: "Is the process enabled?",
+			},
+			"customization_type": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Indicates the type of customization on this process. System Process is default process. Inherited Process is modified process that was System process before.",
+			},
+			"projects": {
+				Type: schema.TypeSet,
```
magodo: Can we change this to TypeList instead?

azuredevops/internal/service/workitemtrackingprocess/data_process.go
```diff
@@ -0,0 +1,154 @@
...
+		return diag.Errorf(" Getting process with id: %s. Error: %+v", processId, err)
+	}
+
+	d.Set("name", process.Name)
+	d.Set("description", process.Description)
+	d.Set("parent_process_type_id", process.ParentProcessTypeId.String())
+	d.Set("reference_name", process.ReferenceName)
+	d.Set("is_default", process.IsDefault)
+	d.Set("is_enabled", process.IsEnabled)
+	d.Set("customization_type", string(*process.CustomizationType))
```
magodo: These also need nil checks.

azuredevops/internal/service/workitemtrackingprocess/data_process.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if process.Projects != nil {
+		var projects []map[string]any
+		for _, p := range *process.Projects {
+			project := map[string]any{
+				"id":          p.Id.String(),
+				"description": p.Description,
+				"name":        p.Name,
+				"url":         p.Url,
+			}
```
magodo, line 165: Nil check needed.

azuredevops/internal/service/workitemtrackingprocess/data_processes.go
```diff
@@ -0,0 +1,195 @@
...
+							Computed:    true,
+							Description: "Is the process enabled?",
+						},
+						"customization_type": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "Indicates the type of customization on this process. System Process is default process. Inherited Process is modified process that was System process before.",
+						},
+						"projects": {
+							Type: schema.TypeSet,
```
magodo: This wants to be TypeList

azuredevops/internal/service/workitemtrackingprocess/data_processes.go
```diff
@@ -0,0 +1,195 @@
...
+		Schema: map[string]*schema.Schema{
+			"expand": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "none",
+				ValidateFunc: validation.StringInSlice([]string{"none", "projects"}, false),
+				Description:  "Specifies the expand option when getting the processes.",
+			},
+			"processes": {
+				Type:        schema.TypeSet,
```
magodo: This wants TypeList

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_process_test.go
Fresa: that make sense. Alright, let's leave it as is then.

azuredevops/internal/service/workitemtrackingprocess/data_process.go
Fresa: I can do that. I'm curious though; why would we want it to be a TypeList and not a TypeSet? 

Functional wise they are equivalent when defined on a data source AFAIK, except how one would consume them via HCL. I semantically interpret projects as unordered though, hence why I picked TypeSet.

azuredevops/internal/service/workitemtrackingprocess/data_process.go
Fresa: totally forgot, good catch!

azuredevops/internal/service/workitemtrackingprocess/data_process.go
magodo: For a writable attribute, we prefer to use TypeList as historically TypeSet has a few bugs, as long as the API keeps the order.

For a read-only attribute, we shall stick to using TypeList as there is no beneift of using TypeSet, as well as there was/is an issue to reference element of a TypeSet.

Fresa: cool, thanks the explanation! I've changed all TypeSet attributes to TypeList

azuredevops/internal/service/workitemtrackingprocess/data_processes.go
Fresa: done

azuredevops/internal/service/workitemtrackingprocess/data_processes.go
Fresa: done

azuredevops/internal/service/workitemtrackingprocess/resource_process.go
Fresa: I've updated this again, I think I might have misunderstood what you meant first. It now avoids panicing on dereferencing nil pointers, but it sets the attribute to nil if the API says it should be.

magodo:
@Fresa Thanks for all the work! It LGTM now! Would you please:

1. Solve the CI failures
2. Provide a screenshot/terminal run of the acceptance test to prove everything works

Then I'll go ahead to merge this PR 👍

Fresa:
@magodo 
image

magodo (approved):

Issue #1031, opened 2024-04-19
Azure DevOps Service Connection: Share service connection with other projects


mateustanaka:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Share and unshare a service connection with other projects.

### New or Affected Resource(s)

* `azuredevops_serviceendpoint_project_permissions`

> `serviceendpoint_id` - (Required) Endpoint Id of the service connection to share
> `project_reference` - (Required) A list of project_reference block as defined below. Objects describing which projects the service connection will be shared
> An `project_reference` block supports the following:
> - `project_id` - (Required) Project id which service endpoint will be shared
> - `service_endpoint_name` - (Optional) Name for service connection in the shared project. Default keep the same name
> - `description` - (Optional) Description for service connection in the shared project. Default keep the same description

### Potential Terraform Configuration

```hcl
resource "azuredevops_serviceendpoint_project_permissions" "example-share" {
  serviceendpoint_id = azuredevops_serviceendpoint_azurerm.example.id

  project_reference = {
    project_id            = azuredevops_project.example_one.id
    service_endpoint_name = "service-connection-shared"
    description           = "Service Connection Shared by Terraform - Cluster One"
  }

  project_reference = {
    project_id            = azuredevops_project.example_two.id
    service_endpoint_name = "service-connection-shared"
    description           = "Service Connection Shared by Terraform - Cluster Two"
  }
}
```

### References

* [Azure DevOps Services REST API](https://learn.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpoints/share-service-endpoint?view=azure-devops-rest-7.1&tabs=HTTP)

mouismail:
👋 I can work on this one using the following plan:

### Introduce [resources_serviceendpoint_permissions.go](https://github.com/microsoft/terraform-provider-azuredevops/tree/main/azuredevops/internal/service/serviceendpoint) 
  - [ ] Add support for `azuredevops_serviceendpoint_project_permissions` resource
  - [ ] Implement CRUD operations for the resource
  - [ ] Handle `project_reference` blocks within the resource
[...]

PR #1455, opened 2025-11-25
Adding the option to share an endpoint to an other project #1031

dennismdejong:
## All Submissions:

* [ x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ x] I have updated the documentation accordingly.
* [ x] I have added tests to cover my changes.
* [x ] All new and existing tests passed.
* [ x] My code follows the code style of this project.
* [ x] I ran lint checks locally prior to submission.
* [ x] Have you checked to ensure there aren't other open PRs for the same update/change?

Issue Number: #1031

- [ ] Yes
- [ x] No

- [ ] Yes
- [ x] No


website/docs/r/workitemquery.html.markdown
```diff
@@ -112,12 +112,12 @@ resource "azuredevops_workitemquery_permissions" "query_permissions" {
     azuredevops_workitemquery_folder.team_folder.name,
```
magodo, line 112: Presumably this file is added by accident?

azuredevops/internal/service/serviceendpoint/commons.go
```diff
@@ -464,3 +505,37 @@ func checkServiceConnection(endpoint *serviceendpoint.ServiceEndpoint) error {
...
+		if ref.ProjectReference != nil {
+			pRef := make(map[string]interface{})
+			if ref.ProjectReference.Id != nil {
+				pRef["id"] = ref.ProjectReference.Id.String()
+			}
+			if ref.ProjectReference.Name != nil {
+				pRef["name"] = *ref.ProjectReference.Name
+			}
+			item["project_reference"] = []interface{}{pRef}
+		}
```
magodo: Can we do:


```suggestion
		if ref := ref.ProjectReference; ref != nil {
			pRef := make(map[string]interface{})
			if id := ref.Id; id != nil {
				pRef["id"] = id.String()
			}
			if name: ref.Name; name != nil {
				pRef["name"] = *name
			}
			item["project_reference"] = []interface{}{pRef}
		}
```

azuredevops/internal/service/serviceendpoint/commons.go
```diff
@@ -362,6 +371,38 @@ func dataSourceGenBaseSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Computed: true,
 		},
+		"service_endpoint_project_references": {
```
magodo: Since this affects all those service endpoint data sources, could you please also reflect that in their documents respectively?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+			},
+			"project_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+				Description:  "The ID of the project where the service endpoint is created (Source Project).",
+			},
+			"project_reference": {
+				Type:     schema.TypeSet,
```
magodo: Does it have to be a set, or a list can also work (assuming the API will keep the order)? We shall avoid using set given it has a few inherent bugs.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func resourceCreateOrUpdateServiceEndpointProjectPermissions(ctx context.Context, d *schema.ResourceData, m interface{}) diag.Diagnostics {
+	clients := m.(*client.AggregatedClient)
+
+	endpointID, err := uuid.Parse(d.Get("service_endpoint_id").(string))
+	if err != nil {
+		return diag.Errorf("invalid service_endpoint_id: %v", err)
+	}
+	targetProjectID, err := uuid.Parse(d.Get("project_id").(string))
```
magodo: shouldn't this be sourceProjectID instead?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+	if err != nil {
+		return diag.Errorf("invalid project_id: %v", err)
+	}
+
+	serviceEndpoint, err := clients.ServiceEndpointClient.GetServiceEndpointDetails(
+		clients.Ctx,
```
magodo:
```suggestion
		ctx,
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+			Project:    converter.String(targetProjectID.String()),
+		},
+	)
+	if err != nil {
+		return diag.Errorf("Error finding service endpoint: %+v", err)
+	}
+
+	oldSet, newSet := d.GetChange("project_reference")
+
+	projectsToRemove := make(map[string]bool)
```
magodo: This seems not necessary. Since the Terraform config is regarded as the expected state, we shall simply just get the planned project list, which are regarded as `projectsToUpsert`. Then iterate the `serviceEndpoint.ServiceEndpointProjectReferences` and mark the ones to remove. Otherwise, those untracked reference will not be deleted since they are not in the state (i.e. `oldSet`)?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+				Id: &targetProjectID,
+			},
+			Name:        converter.String(name),
+			Description: converter.String(desc),
+		})
+	}
+
+	serviceEndpoint.ServiceEndpointProjectReferences = &newReferences
+	_, err = clients.ServiceEndpointClient.UpdateServiceEndpoint(
+		clients.Ctx,
```
magodo:
```suggestion
		ctx,
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+			ProjectReference: &serviceendpoint.ProjectReference{
+				Id: &targetProjectID,
+			},
+			Name:        converter.String(name),
+			Description: converter.String(desc),
+		})
+	}
+
+	serviceEndpoint.ServiceEndpointProjectReferences = &newReferences
+	_, err = clients.ServiceEndpointClient.UpdateServiceEndpoint(
```
magodo: Per my test, the PATCH is incrementally adding the new reference, but won't replace the whole reference list. This means it won't remove the existing reference that doesn't exist in `projectsToUpsert`.

Therefore, you'll need to explicitly delete the serviceendpoint from the corresponding projects, by iterating a `projectsToDelete` array/map.

Furthermore, the name (i.e. `service_endpoint_name`) of the shared endpoint can't be updated with this API as well, in which case the API will raise the following error:

> TF400898: An Internal Error Occurred

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+			if _, found := projectsToDelete[pid]; !found {
+				newRefs = append(newRefs, ref)
+			}
+		}
+	}
+
+	serviceEndpoint.ServiceEndpointProjectReferences = &newRefs
+
+	_, err = clients.ServiceEndpointClient.UpdateServiceEndpoint(
+		clients.Ctx,
```
magodo:
```suggestion
		ctx,
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+
+	endpointIDStr := d.Get("service_endpoint_id").(string)
+	targetProjectIDStr := d.Get("project_id").(string)
+	endpointID, err := uuid.Parse(endpointIDStr)
+	if err != nil {
+		return diag.Errorf("f%s", err)
+	}
+
+	serviceEndpoint, err := clients.ServiceEndpointClient.GetServiceEndpointDetails(
+		clients.Ctx,
```
magodo:
```suggestion
		ctx,
```

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+
+			if _, found := projectsToDelete[pid]; !found {
+				newRefs = append(newRefs, ref)
+			}
+		}
+	}
+
+	serviceEndpoint.ServiceEndpointProjectReferences = &newRefs
+
+	_, err = clients.ServiceEndpointClient.UpdateServiceEndpoint(
```
magodo: This might also not work as expected, instead we shall remove the endpoint from the target project instead.

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,281 @@
...
+		clients.Ctx,
+		serviceendpoint.UpdateServiceEndpointArgs{
+			EndpointId: &endpointID,
+			Endpoint:   serviceEndpoint,
+		},
+	)
+
+	return diag.Errorf("f%s", err)
```
magodo: Is this on purpose to always return an error?

azuredevops/internal/acceptancetests/resource_serviceendpoint_project_permissions_test.go
```diff
@@ -0,0 +1,104 @@
...
+	serviceEndpointName := testutils.GenerateResourceName()
+
+	tfNode := "azuredevops_serviceendpoint_project_permissions.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testutils.PreCheck(t, nil) },
+		Providers:    testutils.GetProviders(),
+		CheckDestroy: testutils.CheckProjectDestroyed,
+		Steps: []resource.TestStep{
+			// Stap 1: Initieel delen met Project 2
```
magodo: Can we translate this (and the other comments) to English?

magodo (changes requested):
Thanks for this PR!

mmalyakov:
@magodo, @dennismdejong 

Any progress on this one?

kaiyuanlim-adh:
Would be nice to have this soon. Looking forward to it.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_azurecr.go
```diff
@@ -80,7 +80,7 @@ func dataSourceServiceEndpointAzureCRRead(d *schema.ResourceData, m interface{})
 		if err = checkServiceConnection(serviceEndpoint); err != nil {
 			return err
 		}
-		doBaseFlattening(d, serviceEndpoint)
+		doBaseFlatteningForDataSource(d, serviceEndpoint)
```
magodo: Any reason to only change this data source?

azuredevops/internal/service/serviceendpoint/resource_serviceendpoint_project_permissions.go
```diff
@@ -0,0 +1,253 @@
...
+			"project_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.IsUUID,
+				Description:  "The ID of the project where the service endpoint is created (Source Project).",
+			},
+			"project_reference": {
+				Type:     schema.TypeList,
+				Optional: true,
```
magodo: Should we mark this as required as otherwise it is the same state as it is deleted as a terraform resource?

magodo (changes requested):
Thanks for this PR!

website/docs/d/serviceendpoint_azurecr.html.markdown
```diff
@@ -64,6 +64,21 @@ In addition to the Arguments list above - the following Attributes are exported:
```
magodo, line 64: Since you've deleted the schema change for data source, can we also revert these changes to the data source documents?

magodo (changes requested):

musukvl:
Hello, 
Please complete the Pull Request. 
The "sharing endpoint to an other project" is an important feature in scope the latest changes in Terraform extension of microsoft. Right now we have "Client secret authentication is not secure and will be deprecated in the next major version of this task. Please use Workload identity federation authentication instead" warning for all our pipelines. 
The best way for us to share service connection with federated credentials across the projects, to avoid configuration of app registration for each service connection.

PR #1460, opened 2025-11-28
New resource: `azuredevops_workitemtrackingprocess_workitemtype`; New data sources: `azuredevops_workitemtrackingprocess_workitemtype`, `azuredevops_workitemtrackingprocess_workitemtypes`

Fresa:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [ ] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## What about the current behavior has changed?
~~NOTE: This is based on [PR 1454](https://github.com/microsoft/terraform-provider-azuredevops/pull/1454), so it should be merged first.~~

Introduces the resource `azuredevops_workitemtrackingprocess_workitemtype` and corresponding data sources `azuredevops_workitemtrackingprocess_workitemtype` and `azuredevops_workitemtrackingprocess_workitemtypes`.

The complex expandable structure of a work item type is not included except references for the layout, which is tightly coupled to a work item type. I think it will be simpler to deal with the other structures as separate resources.

Issue Number: [1446](https://github.com/microsoft/terraform-provider-azuredevops/issues/1446)

## Does this introduce a change to `go.mod`, `go.sum` or `vendor/`?

- [ ] Yes
- [x] No

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Any relevant logs, error output, etc?
N/A

## Other information

There are a couple of unrelated unit tests failing, but they are failing on main to.

PR #1462, opened 2025-12-04
Add work item type groups

Fresa:
## All Submissions:

Add [groups](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/groups?view=azure-devops-rest-7.1) to work item type layouts (sections in pages).

~~This is based on [PR 1460](https://github.com/microsoft/terraform-provider-azuredevops/pull/1460) which probably should be merged first.~~

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


Fresa:
@magodo here's another one, when ever you have time!

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_group_test.go
```diff
@@ -0,0 +1,224 @@
+//go:build (all || resource_workitemtrackingprocess || resource_workitemtrackingprocess_group) && !exclude_resource_workitemtrackingprocess
+// +build all resource_workitemtrackingprocess resource_workitemtrackingprocess_group
+// +build !exclude_resource_workitemtrackingprocess
```
magodo: Can we remove these which are not needed any more?

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
```diff
@@ -0,0 +1,290 @@
...
+	createdGroup, err := clients.WorkItemTrackingProcessClient.AddGroup(ctx, args)
+	if err != nil {
+		return diag.Errorf(" Creating group. Error %+v", err)
+	}
+
+	if createdGroup.Id == nil {
+		return diag.Errorf(" Created group has no ID")
+	}
+
+	d.SetId(*createdGroup.Id)
```
magodo, line 127: Can we set the same id format as is used during import here?

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
```diff
@@ -0,0 +1,290 @@
...
+	return readResourceGroup(ctx, d, m)
+}
+
+func readResourceGroup(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	clients := m.(*client.AggregatedClient)
+
+	groupId := d.Id()
+	processId := d.Get("process_id").(string)
+	witRefName := d.Get("work_item_type_reference_name").(string)
```
magodo, line 136: The id is expected to contain these values and shall be extracted from the id instead of the state, which might be missing during import.

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
```diff
@@ -0,0 +1,290 @@
...
+
+	// Get the work item type with layout expanded
+	getWorkItemTypeArgs := workitemtrackingprocess.GetProcessWorkItemTypeArgs{
```
magodo, line 139: Seems each wit can only reside in one group, also the group information is held in the wit response (expanded). Does it make sense to make the group related information as a set of properties in the wit resource instead?

magodo (changes requested):

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_group_test.go
Fresa: done

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
Fresa: it is the same format? The id is the [group id](https://github.com/microsoft/terraform-provider-azuredevops/blob/edfd436e681e0e01f89d0ae290dda04e307cd2ce/azuredevops/internal/service/workitemtrackingprocess/resource_group.go#L83)

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
Fresa: the import function [parses the import identifier](https://github.com/microsoft/terraform-provider-azuredevops/blob/edfd436e681e0e01f89d0ae290dda04e307cd2ce/azuredevops/internal/service/workitemtrackingprocess/resource_group.go#L89-L93) and sets the appropriate resource attributes. The acceptance tests have import steps that should verify that this works correctly, or am I missing something?

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
Fresa: WIT's doesn't reside in groups, it defines a [layout](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/customize-process-form?view=azure-devops) that has groups (within sections and pages). A group can group controls. The layout structure is [pretty nested and complex](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/layout/get?view=azure-devops-rest-7.1&tabs=HTTP#get-the-form-layout) and can be composed by many entities. I chose to separate all these entities (groups, controls, pages) into separate resources to make them simpler to handle, and to align with the devops API that also defines them as separate entities, [groups](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/groups?view=azure-devops-rest-7.1), [controls](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/controls?view=azure-devops-rest-7.1), [pages](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/pages?view=azure-devops-rest-7.1).

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
magodo: Per the example in https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/groups/add?view=azure-devops-rest-7.2&tabs=HTTP, the `createGroup.Id` here is just an uuid?

magodo (approved):

magodo:
LGTM! Thanks!

PR #1449, opened 2025-11-18
Implement generic permission resource with token generation helper data source.

aklakina:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## What about the current behavior has changed?

This PR introduces generic security management capabilities for Azure DevOps, providing low-level access to the security namespace system. The implementation includes:

**Data Sources:**
- `azuredevops_security_namespaces` - Discovers available security namespaces and their permissions across Azure DevOps (Collection, Project, Git Repositories, Analytics, etc.)
- `azuredevops_security_namespace_token` - Generates security tokens for specific resources, with support for discovering required identifiers per namespace. This data source is NOT dynamic. Any change to expected token format or value on any namespace WILL break it. These changes should not happen frequently though. 

**Resource:**
- `azuredevops_security_permissions` - Generic permissions management across any security namespace, with validation of permission names and support for both replace and merge strategies

These resources complement the existing type-specific permission resources (project, git, build definitions) by exposing the underlying security model, enabling users to manage permissions for any Azure DevOps resource type, including those without dedicated Terraform resources.

Issue Number:
- #340
- #354
- #404
- #521
- #850
- #998
- #1023
- #1177
- #1217
- #1344

## Does this introduce a change to `go.mod`, `go.sum` or `vendor/`?

- [ ] Yes
- [x] No

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Any relevant logs, error output, etc?

(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Issue #1344, opened 2025-04-03
azuredevops_project_permissions missing documentation for Analytics permissions


tlfzhylj:
In the resource [`azuredevops_project_permissions`](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/project_permissions) I can't find the correct keys for configuring permissions for Analytics.

<img width="613" alt="Image" src="https://github.com/user-attachments/assets/d8c29b28-06e5-4144-acc3-1e7514768ac3" />

xuzhang3:
Permission token has been changed when manage **Analytics** permissions.  `$/Shared/<project ID>`

tlfzhylj:
I'm sorry, but I don't understand if this was ment for me...

xuzhang3:
@tlfzhyl Sorry for the misleading messages. My last reply was to find the root cause of this issue. If anyone wants to fix this issue, there should be some useful information.

Issue #1217, opened 2024-11-07
`azuredevops_git_permissions` fails to apply on Azure DevOps Server 2022


paddymorgan84:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version
Terraform: v1.9.8
Azure DevOps Provider: 1.4.0

### Affected Resource(s)

* `azuredevops_git_permissions`

### Terraform Configuration Files

```hcl
data "azuredevops_project" "Project" {
  name = var.project_name
}

data "azuredevops_identity_group" "contributors" {
  project_id = data.azuredevops_project.MedisoftEMR.id
  name       = "[${var.project_name}]\\Contributors"
}

resource "azuredevops_git_permissions" "contributors-permissions" {
  project_id    = data.azuredevops_project.Project.id
  repository_id = var.repository_id
  principal     = data.azuredevops_identity_group.contributors.id
  permissions = {
    PolicyExempt            = "Deny"
    PullRequestBypassPolicy = "Deny"
    GenericContribute       = "Allow"
    PullRequestContribute   = "Allow"
    CreateBranch            = "Allow"
    CreateTag               = "Allow"
    DeleteRepository        = "Deny"
    EditPolicies            = "Deny"
    ForcePush               = "Deny"
    ManageNote              = "Allow"
    ManagePermissions       = "Deny"
    GenericRead             = "Allow"
    RemoveOthersLocks       = "Deny"
    RenameRepository        = "Deny"
  }
}
```

### Expected Behavior

Permissions should have been set for the contributors project group at the repository level.

### Actual Behavior

An error is thrown:

```
 Error: The string must have at least one character.
│ Parameter name: descriptors element.IdentityType
│ 
│   with module.repository["monitoring"].module.permissions.azuredevops_git_permissions.contributors-permissions,
│   on modules/repository/modules/permissions/main.tf line 10, in resource "azuredevops_git_permissions" "contributors-permissions":
│   10: resource "azuredevops_git_permissions" "contributors-permissions" {
```

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

[...]

Issue #1177, opened 2024-10-03
azuredevops_build_folder_permissions


huczas:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/hashicorp/azuread v2.53.1
+ provider registry.terraform.io/hashicorp/azurerm v4.1.0
+ provider registry.terraform.io/hashicorp/random v3.6.2
+ provider registry.terraform.io/hashicorp/time v0.12.0
+ provider registry.terraform.io/microsoft/azuredevops v1.3.0

### Affected Resource(s)

* azuredevops_build_folder_permissions

### Terraform Configuration Files

n/a

### Debug Output

n/a

### Panic Output

n/a

### Expected Behavior

be able to configure "Edit queue build configuration" allow/deny

### Actual Behavior

There is no way to set this permission in a module.

### Steps to Reproduce

n/a

1. `terraform apply`

ic-vbiradar:
+1

ic-jrekkanchat:
+1

ic-pmandadi:
+1

IC-Abhimitra:
+1

mahmed003:
+1

shahjebansari:
+1

IC-sbiradar:
+1

hkarthik7:
Hiya, The resource "**azuredevops_build_folder_permissions**" accepts all the actions of build security namespace ID **33344d9c-fc72-4d6f-aba5-fa317101a7e9**. Please see the below keys that can be used inside permission block. I've tried this and it is working.

**EditPipelineQueueConfigurationPermission** is key that sets permission for **Edit queue build configuration**.

```terraform
terraform {
  required_providers {
    azuredevops = {
      source  = "microsoft/azuredevops"
      version = "1.8.1"
    }
  }
}

provider "azuredevops" {
  
}

data "azuredevops_project" "project" {
  name = "sample-project"
}

data "azuredevops_group" "group_build_admins" {
[...]

PR #1471, opened 2025-12-22
Add new data sources: `azuredevops_security_namespaces` and `azuredevops_security_namespace`

aklakina:
## All Submissions:

This is PR#1 of 3 PRs introducing generic security management capabilities for Azure DevOps, providing low-level access to the security namespace system. (see #1449 for parent pr)

**Data Sources:**
- `azuredevops_security_namespaces` - Discovers available security namespaces and their permissions across Azure DevOps (Collection, Project, Git Repositories, Analytics, etc.)

This data source exposes the [security namespace](https://learn.microsoft.com/en-us/rest/api/azure/devops/security/security-namespaces) api endpoint.

Issue Number:
- #340
- #354
- #404
- #521
- #850
- #998
- #1023
- #1177
- #1217
- #1344

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
+++ b/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
@@ -1,9 +1,12 @@
 package acceptancetests
 
 import (
+	"fmt"
+	"strconv"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
 )
 
@@ -17,6 +20,19 @@ func TestAccDataSecurityNamespaces_basic(t *testing.T) {
 				Config: hclDataSecurityNamespaces_basic(),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttrSet(tfNode, "namespaces.#"),
+					func(s *terraform.State) error {
+						rs, ok := s.RootModule().Resources[tfNode]
+						if !ok {
+							return fmt.Errorf("Not found: %s", tfNode)
+						}
+
+						namespaces := rs.Primary.Attributes["namespaces.#"]
+						count, _ := strconv.Atoi(namespaces)
+						if count == 0 {
+							return fmt.Errorf("Security namespaces list is empty")
+						}
+						return nil
+					},
 				),
 			},
 		},
--- a/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
+++ b/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
@@ -1,0 +1,30 @@
+package acceptancetests
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
+)
+
+func TestAccDataSecurityNamespaces_basic(t *testing.T) {
+	tfNode := "data.azuredevops_security_namespaces.test"
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testutils.PreCheck(t, nil) },
+		ProviderFactories: testutils.GetProviderFactories(),
+		Steps: []resource.TestStep{
+			{
+				Config: hclDataSecurityNamespaces_basic(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(tfNode, "namespaces.#"),
+				),
+			},
+		},
+	})
+}
+
+func hclDataSecurityNamespaces_basic() string {
+	return `
+data "azuredevops_security_namespaces" "test" {}
+`
+}
--- a/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
+++ b/azuredevops/internal/acceptancetests/data_security_namespaces_test.go
@@ -27,8 +27,8 @@ func TestAccDataSecurityNamespaces_basic(t *testing.T) {
 						}
 
 						namespaces := rs.Primary.Attributes["namespaces.#"]
-						count, _ := strconv.Atoi(namespaces)
-						if count == 0 {
+						count, err := strconv.Atoi(namespaces)
+						if count == 0 || err != nil {
 							return fmt.Errorf("Security namespaces list is empty")
 						}
 						return nil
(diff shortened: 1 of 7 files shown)


website/docs/d/security_namespaces.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+
+* `display_name` - The display name of the security namespace.
+
+* `description` - The description of the security namespace (if available).
+
+* `actions` - A set of available actions (permissions) in this namespace. Each `action` block exports the following:
+  * `name` - The name of the action/permission.
+  * `display_name` - The display name of the action/permission.
+  * `bit` - The bit value for this permission (used in permission calculations).
+  * `namespace_id` - The namespace ID this action belongs to.
```
magodo: This seems misleading. At the first glance, it seems to be redundant with the top level `namespace_id`, while it actually returns `00000000-0000-0000-0000-000000000000` in my case (when querying for a project namespace).

website/docs/d/security_namespaces.html.markdown
```diff
@@ -0,0 +1,140 @@
...
+      display_name = action.display_name
+      bit          = action.bit
+    }
+  }
+}
+```
+
+## Argument Reference
+
+This data source does not require any arguments.
```
magodo, line 54: Does it make sense to also implement a `azuredevops_security_namespace` data source that queries for a single namespace, by a user-friendly name from a pre-defined list (can reusing its `name` attribute)?

azuredevops/internal/service/security/data_security_namespaces.go
```diff
@@ -0,0 +1,157 @@
...
+// DataSecurityNamespaces schema and implementation for security namespaces data source
+func DataSecurityNamespaces() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSecurityNamespacesRead,
+		Timeouts: &schema.ResourceTimeout{
+			Read: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: map[string]*schema.Schema{
+			"namespaces": {
+				Type:     schema.TypeSet,
```
magodo: Can we change this to `TypeList` as it is Computed only?

azuredevops/internal/service/security/data_security_namespaces.go
```diff
@@ -0,0 +1,157 @@
...
+				Computed: true,
+				Set:      getSecurityNamespaceHash,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"namespace_id": {
```
magodo: Consider to rename it to `id`.

azuredevops/internal/service/security/data_security_namespaces.go
```diff
@@ -0,0 +1,157 @@
...
+							Computed:    true,
+							Description: "The display name of the security namespace",
+						},
+						"description": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The description of the security namespace",
+						},
+						"actions": {
+							Type:        schema.TypeSet,
```
magodo: Please change to TypeList.

azuredevops/internal/service/security/data_security_namespaces.go
```diff
@@ -0,0 +1,157 @@
...
+										Type:        schema.TypeString,
+										Computed:    true,
+										Description: "The display name of the action/permission",
+									},
+									"bit": {
+										Type:        schema.TypeInt,
+										Computed:    true,
+										Description: "The bit value for this permission",
+									},
+									"namespace_id": {
```
magodo: Consider to remove this if not useful.

azuredevops/internal/service/security/data_security_namespaces.go
```diff
@@ -0,0 +1,157 @@
...
+			namespace["namespace_id"] = ns.NamespaceId.String()
+		}
+		if ns.Name != nil {
+			namespace["name"] = *ns.Name
+		}
+		if ns.DisplayName != nil {
+			namespace["display_name"] = *ns.DisplayName
+		}
+		// Note: SecurityNamespaceDescription does not have a Description field
+		namespace["description"] = ""
```
magodo: Can we delete this then?

azuredevops/internal/acceptancetests/data_security_namespaces_test.go
```diff
@@ -0,0 +1,30 @@
...
+func TestAccDataSecurityNamespaces_basic(t *testing.T) {
+	tfNode := "data.azuredevops_security_namespaces.test"
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testutils.PreCheck(t, nil) },
+		ProviderFactories: testutils.GetProviderFactories(),
+		Steps: []resource.TestStep{
+			{
+				Config: hclDataSecurityNamespaces_basic(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(tfNode, "namespaces.#"),
```
magodo, line 22: Can we check it is not equals to 0?

magodo (changes requested):
Thanks for this PR!

website/docs/d/security_namespaces.html.markdown
aklakina: The API does not allow querying by name, so the request would be the same. If I were to implement such thing, then maybe I would create a singleton cache and then just return from memory. Should I add it to this pr?

azuredevops/internal/acceptancetests/data_security_namespaces_test.go
aklakina: Added the check.

azuredevops/internal/service/security/data_security_namespaces.go
aklakina: Deleted and removed the property.

azuredevops/internal/service/security/data_security_namespaces.go
aklakina: Removed the namespace_id subproperty.

website/docs/d/security_namespaces.html.markdown
aklakina: Weird, it should be the same as the top-level id. But I removed the property instead as it is a duplicate

aklakina:
Hi @magodo

I've implemented the requested changes, can you please have a look at it?

website/docs/d/security_namespaces.html.markdown
magodo: Yes, that would be a better UX for the customers. The cachine might be unnecessary as there is a rare case that users will specify multiple data sources for the same namespace.

magodo:
@aklakina Thanks for the updates! It LGTM! Please fix the CI and attach a run of the acctests, I'll be happy to merge it!

aklakina:
image
Here is the output of the acceptance tests. @magodo

Also fixed the linter issues.

aklakina:
Hey @magodo

Can this be merged then?

magodo (approved):

Issue #1023, opened 2024-04-10
Missing resource to set Analytics and Analytics views permissions


alborworld:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently, there is no resource to set the permissions for the Analytics and Analytics views. 

### New or Affected Resource(s)

* azuredevops_analyticsviews_permissions

### Potential Terraform Configuration

```hcl
resource "azuredevops_project" "example" {
  name               = "Example Project"
  work_item_template = "Agile"
  version_control    = "Git"
  visibility         = "private"
  description        = "Managed by Terraform"
}

data "azuredevops_group" "example-project-readers" {
  project_id = azuredevops_project.example.id
  name       = "Readers"
}

resource "azuredevops_analyticsviews_permissions" "example" {
  project_id = azuredevops_project.example.id
  principal  = data.azuredevops_group.example-project-readers.id

  permissions = {
    Read    = "Deny"
    Edit      = "Allow"
    Delete   = "Deny"
    Execute    = "Allow"
    ManagePermissions = "Allow"
  }
}
```

The *principal* can be a member, a team, a group. 

If the permissions can be set at project-level (to be checked) then *principal* should be **optional**.

### References

AnalyticsViews namespace: https://learn.microsoft.com/en-us/azure/devops/report/powerbi/analytics-security?view=azure-devops-2022&tabs=current-page

Set permissions to access Analytics and Analytics views: https://learn.microsoft.com/en-us/azure/devops/report/powerbi/analytics-security?view=azure-devops-2022&tabs=current-page

* #0000

Issue #998, opened 2024-03-11
Pipeline Permissions


philiplorenz-code:
Does this provider provide any possibility to enable a Custom Group, that was build with this provider, for creating pipelines on a specific project?

xuzhang3:
@philiplorenz-code  you can manage the [group](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/group), [pipeline](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/build_definition) and [project](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/project) by this provider.

philiplorenz-code:
> @philiplorenz-code you can manage the [group](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/group), [pipeline](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/build_definition) and [project](https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/project) by this provider.

thanks, but that was not my question :) I enabled the pipeline-feature for a project. My question ist, if it is possible to authorize a _custom group_ to create pipelines within this project :)

xuzhang3:
@philiplorenz-code Which resource can be managed depends on the permission of the PAT not group.

PascalTurbo:
Hey, we are facing the same issue. There is no documentation on how to grant "Create Pipeline Permissions" to a custom group.

Issue #850, opened 2023-08-02
azuredevops_serviceendpoint_permissions does not correctly remove permissions when partially removed from configuration


liyaoz:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version
Terraform 0.14.11
Azure DevOps provider 0.6.0

### Affected Resource(s)

* `azuredevops_serviceendpoint_permissions`

### Terraform Configuration Files

This code was applied, and users started complaining losing access for being both in group xxx and endpoint admin groups.
```hcl
resource "azuredevops_serviceendpoint_permissions" "service_endpoint_permissions" {
  project_id = azuredevops_project.this.id
  principal  = azuredevops_group.xxx.id
  permissions = {
    Use               = "allow"
    Administer        = "deny"
    Create            = "deny"
    ViewAuthorization = "allow"
    ViewEndpoint      = "allow"
  }
}
```
This was applied in attempt to remove explicit "deny"s:
```hcl
resource "azuredevops_serviceendpoint_permissions" "service_endpoint_permissions" {
  project_id = azuredevops_project.this.id
  principal  = azuredevops_group.xxx.id
  permissions = {
    Use               = "allow"
    ViewAuthorization = "allow"
    ViewEndpoint      = "allow"
  }
}
```
It fixed for new groups but didn't work for existing groups and users.
Finally this helped:
```hcl
resource "azuredevops_serviceendpoint_permissions" "service_endpoint_permissions" {
  project_id = azuredevops_project.this.id
  principal  = azuredevops_group.xxx.id
  permissions = {
    Use               = "allow"
    Administer        = "notset"
    Create            = "notset"
    ViewAuthorization = "allow"
    ViewEndpoint      = "allow"
  }
}
```

### Expected Behavior
This should either error out, or set the unset permissions to 'notset' as default.
```hcl
  permissions = {
    Use               = "allow"
    ViewAuthorization = "allow"
    ViewEndpoint      = "allow"
  }
```
Also the documentation didn't mention what values are allowed in this object. I found "notset" in the source code from tests.

### Actual Behavior
[...]

Issue #521, opened 2022-01-19
configure project-level build permissions


RodrigoGroener:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently it's only possible to define build permissions on object level (build definition).

Make `build_definition_id` argument optional, so that it's possible to define the project-level build permissions.

This would correspond the `azuredevops_git_permissions` where it's possible to define the permissions on project level, repository level and branch level by defining an additional optional parameter.

### New or Affected Resource(s)

* azuredevops_build_definition_permissions

### Potential Terraform Configuration

```hcl
resource "azuredevops_project" "project" {
  name       = "Sample Project"
  work_item_template = "Agile"
  version_control    = "Git"
  visibility         = "private"
  description        = "Managed by Terraform"
}

data "azuredevops_group" "project-readers" {
    project_id = azuredevops_project.project.id
    name       = "Readers"
}

resource "azuredevops_build_definition_permissions" "permissions" {
    project_id  = azuredevops_project.project.id
    principal   = data.azuredevops_group.project-readers.id

    permissions = {
      ViewBuilds       = "Allow"
      EditBuildQuality = "Deny"
      DeleteBuilds     = "Deny"
      StopBuilds       = "Allow"
    }
}
```

### References

* #346 
* This could also be the first step to support the following issues:
  * #354 
  * #353 

* [`Build` Namespace - Azure DevOps permission reference](https://docs.microsoft.com/en-us/azure/devops/organizations/security/namespace-reference?view=azure-devops#object-level-namespaces-and-permissions)

lindhe:
This is a very interesting feature for us too!

racdev:
If the build_folder_permissions resource accepted an empty path (and hence represented the root when no path is specified) then that would allow project level permissions to be managed by that resource  - subject to it generating the correct token when no path is set.

As far as I can tell anyway

magodo:
This is supported by `azuredevops_build_folder_permissions`. In order to set the root folder's permission, just set the `path = "/"`.

With this, I'm gonna close this issue.

PR #1472, opened 2025-12-22
Add new data source: `azuredevops_security_namespace_token` for gener…

aklakina:
## All Submissions:

This is PR#2 of 3 PRs introducing generic security management capabilities for Azure DevOps, providing low-level access to the security namespace system. (see #1449 for parent pr)

**Data Sources:**
- `azuredevops_security_namespace_token` - Generates security tokens for specific resources, with support for discovering required identifiers per namespace. This data source is NOT dynamic. Any change to expected token format or value on any namespace WILL break it. These changes should not happen frequently though. 

This data source is a hardcoded token generator based on examined network traffic for specific permission endpoints. Microsoft does not provide a way to query these tokens.

Issue Number:
- #340
- #354
- #404
- #521
- #850
- #998
- #1023
- #1177
- #1217
- #1344

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


aklakina:
Hey @magodo,

Can you have a look at this pr?

azuredevops/provider.go
```diff
@@ -184,6 +184,7 @@ func Provider() *schema.Provider {
 			"azuredevops_project":                               core.DataProject(),
 			"azuredevops_projects":                              core.DataProjects(),
 			"azuredevops_security_namespace":                    security.DataSecurityNamespace(),
+			"azuredevops_security_namespase_token":              security.DataSecurityNamespaceToken(),
```
magodo:
```suggestion
			"azuredevops_security_namespace_token":              security.DataSecurityNamespaceToken(),
```

azuredevops/provider_test.go
```diff
@@ -163,6 +163,7 @@ func TestProvider_HasChildDataSources(t *testing.T) {
 		"azuredevops_project",
 		"azuredevops_projects",
 		"azuredevops_security_namespace",
+		"azuredevops_security_namespase_token",
```
magodo:
```suggestion
		"azuredevops_security_namespace_token",
```

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+	Template string
+	// RequiredIdentifiers are the identifiers that must be provided
+	RequiredIdentifiers []string
+	// OptionalIdentifiers are the identifiers that may be provided
+	OptionalIdentifiers []string
+	// BuildFunc is an optional custom function to build the token
+	BuildFunc func(identifiers map[string]string) (string, error)
+}
+
+func stringToUTF16LEHex(s string) string {
```
magodo: Any reason not to use converter.EncodeUtf16HexString?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+		OptionalIdentifiers: []string{"repository_id", "ref_name"},
+		BuildFunc: func(identifiers map[string]string) (string, error) {
+			if _, hasRef := identifiers["ref_name"]; hasRef {
+				if _, hasRepo := identifiers["repository_id"]; !hasRepo {
+					return "", fmt.Errorf("ref_name provided without repository_id; ref_name needs it's parent repository_id to build the token")
+				}
+			}
+			if repoID, hasRepo := identifiers["repository_id"]; hasRepo {
+				if refName, hasRef := identifiers["ref_name"]; hasRef {
+					// split refName. The first 2 segments should stay as is, the rest need to be stringToUTF16LEHex
```
magodo, line 183: The existing logic in the git permission resource seems to accept a branch name in any of the following case:
- /refs/heads/foo
- refs/heads/foo
- foo

azuredevops/internal/acceptancetests/data_security_namespace_token_test.go
```diff
@@ -0,0 +1,32 @@
+package acceptancetests
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests/testutils"
+)
+
+func TestAccDataSecurityNamespaceToken_collection(t *testing.T) {
```
magodo, line 21: Can we add tests to cover every supported cases?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	// Convert bytes to hexadecimal string
+	return hex.EncodeToString(bytes)
+}
+
+// namespaceTokenTemplates maps namespace IDs to their token generation templates
+var namespaceTokenTemplates = map[string]TokenTemplate{
+	// Git Repositories namespace
+	"2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87": {
```
magodo: Can we use the `securityhelper.SecurityNamespaceIDValues` values as the keys instead of duplicating these uuids?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/microsoft/azure-devops-go-api/azuredevops/v7/security"
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/client"
+)
+
+// TokenTemplate defines the structure for generating tokens for a namespace
+type TokenTemplate struct {
+	// Template is the format string for the token (if BuildFunc is nil)
+	Template string
```
magodo: Arguably, we can stick using `BuildFunc`.

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+		bytes[i*2+1] = byte(code >> 8) // High byte
+	}
+
+	// Convert bytes to hexadecimal string
+	return hex.EncodeToString(bytes)
+}
+
+// namespaceTokenTemplates maps namespace IDs to their token generation templates
+var namespaceTokenTemplates = map[string]TokenTemplate{
+	// Git Repositories namespace
```
magodo, line 167: Can we extend the comment to list all the possible token formats?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+					// Remove leading/trailing slashes and convert backslashes to forward slashes
+					transformedPath := strings.Trim(strings.ReplaceAll(path, "\\", "/"), "/")
+					return fmt.Sprintf("%s/%s/%s", projectID, transformedPath, buildDefID), nil
+				}
+				return fmt.Sprintf("%s/%s", projectID, buildDefID), nil
+			}
+			if hasPath {
+				// Remove leading/trailing slashes and convert backslashes to forward slashes
+				transformedPath := strings.Trim(strings.ReplaceAll(path, "\\", "/"), "/")
+				return fmt.Sprintf("%s/%s", projectID, transformedPath), nil
```
magodo: Is this a valid token?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+			}
+			return projectID, nil
+		},
+	},
+	// Service Endpoints namespace
```
magodo, line 258: Please extend the comment to elaborate all token formats.

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+		OptionalIdentifiers: []string{"serviceendpoint_id"},
+		BuildFunc: func(identifiers map[string]string) (string, error) {
+			projectID := identifiers["project_id"]
+			if seID, hasSE := identifiers["serviceendpoint_id"]; hasSE {
+				return fmt.Sprintf("endpoints/%s/%s", projectID, seID), nil
+			}
+			return fmt.Sprintf("endpoints/%s", projectID), nil
+		},
+	},
+	// CSS namespace (Areas)
```
magodo, line 266: The existing azuredevops_area_permissions logic seems more complicated. Can you please check?

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+			// Format: vstfs:///Classification/Node/<NodeIdentifier>
+			// For nested: vstfs:///Classification/Node/<RootID>:vstfs:///Classification/Node/<ChildID>
+			if nodeID, hasNode := identifiers["node_id"]; hasNode {
+				return fmt.Sprintf("vstfs:///Classification/Node/%s", nodeID), nil
+			}
+			return "", nil
+		},
+	},
+	// Iteration namespace
+	// Note: Requires API calls to resolve path to node identifiers
```
magodo, line 278: This is also different than the existing logic.

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+		OptionalIdentifiers: []string{"project_id"},
+		BuildFunc: func(identifiers map[string]string) (string, error) {
+			if projectID, hasProject := identifiers["project_id"]; hasProject {
+				return fmt.Sprintf("PublisherSecurity/%s", projectID), nil
+			}
+			return "PublisherSecurity", nil
+		},
+	},
+	// Work Item Query Folders namespace
+	// Note: Requires API calls to resolve path to query IDs
```
magodo, line 311: Same here

azuredevops/internal/service/security/data_security_namespace_token.go
```diff
@@ -0,0 +1,432 @@
...
+		}
+	}
+
+	// Look up the template for this namespace
+	template, exists := namespaceTokenTemplates[namespaceID.String()]
+	if !exists {
+		// For unknown namespaces, provide a basic fallback
+		if projectID, hasProject := identifiers["project_id"]; hasProject {
+			return fmt.Sprintf("$/%s", projectID), nil
+		}
```
magodo: Can we just be explicit to define the buildFunc?

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline, please take a look!

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: I have implemented this based on monitoring the api calls to Azure DevOps during such changes. There it was explicitly `refs/heads|tags/<UTF16LEHEX>`. Should I modify it to accept the same things as the original git permission resource?

azuredevops/internal/acceptancetests/data_security_namespace_token_test.go
aklakina: We could, but the only logic here is the class that either has a template string or a function which is already tested in the unit test. If that logic is tested, I think having it tested multiple times is redundant. Or you aim to test if for some reason the token construction is broken?

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: With this my aim was to provide an easier way to supply more primitive tokens. Obviously the `BuildFunc` can handle everything that the `Template` could. Should I remove this functionality and stick with `BuildFunc` only?

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: Based on the api requests, it should be.

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: I realized that service endpoints uses role assignment feature instead of permission bit based security. I changed this to just return with an error saying one should use role assignment instead.

azuredevops/internal/utils/converter/converter.go
```diff
@@ -149,7 +149,7 @@ func EncodeUtf16HexString(message string) (string, error) {
 	encodedByte := utf16.Encode(runeByte)
 	var sb strings.Builder
 	for i := 0; i < len(encodedByte); i++ {
-		fmt.Fprintf(&sb, "%02x%02x", encodedByte[i], encodedByte[i]>>8)
+		fmt.Fprintf(&sb, "%02x%02x", byte(encodedByte[i]), byte(encodedByte[i]>>8))
```
aklakina, line 152: This function had a data duplication bug. In go fmt the `%02x` means that the minimum width shall be 2 and pads it if it is lower, but does not truncate it if it is larger. For Bytes it is the maximum width and it does actually truncate it.

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: I didn't pay attention to it. Fixed it.

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: I've rebuilt the `BuildFunc` based on the existing logic for all places mentioned. Can you confirm?

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: Instead of this, I just decided that the data source should fail instead if a token generation is not supported.

aklakina:
Hi, @magodo

I fixed most of the things mentioned, can you please review?

azuredevops/internal/service/security/data_security_namespace_token.go
magodo: I think either way works, whilst if you want to keep the current implementation, you'll need to call it out in the document that a ref is expected instead of a branch.

Also, seems the original logic dones't cover other ref types than branch, e.g. tag.

azuredevops/internal/acceptancetests/data_security_namespace_token_test.go
magodo: I'd suggest to also cover them in the acctest regardless the unit test. If you don't have bandwidth to cover every cases, then can we at least to cover each namespace?

azuredevops/internal/service/security/data_security_namespace_token.go
magodo: Yes, I feel that is more consistent.

magodo:
Apart from the above, for the build namespace, can we also use API calls to discover the path from the build definition, instead of from user input?

azuredevops/internal/service/security/data_security_namespace_token.go
aklakina: I've removed the `Template` property.

aklakina:
> Apart from the above, for the build namespace, can we also use API calls to discover the path from the build definition, instead of from user input?

I could add it as an option so that a user may provide a definition id and for that definition, you will get your scope. The problem is if I were to remove the manual path input, then the scope generation would lose the functionality to generate it for folders instead of just pipelines. I think both is required here but it is a good point to add the definition id as a valid input and resolve the scope from that.

azuredevops/internal/acceptancetests/data_security_namespace_token_test.go
aklakina: Added some tests, to cover the namespaces.

aklakina:
@magodo,

I've implemented the changes. I am running the acc test targeting an azure devops server instance, so the build definition can not be tested and hence I added a skip for that. Same for audit log namespace.
Here are the test runs:
image

magodo:
@aklakina Thanks for the update! I've verified that all the test cases are working:

```shell
terraform-provider-azuredevops on  feature/d/security-namespace-token via 🐹 v1.25.5
💤 TF_ACC=1 go test -v -run='TestAccDataSecurityNamespaceToken_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccDataSecurityNamespaceToken_collection
=== PAUSE TestAccDataSecurityNamespaceToken_collection
=== RUN   TestAccDataSecurityNamespaceToken_project
=== PAUSE TestAccDataSecurityNamespaceToken_project
=== RUN   TestAccDataSecurityNamespaceToken_gitRepositories_project
=== PAUSE TestAccDataSecurityNamespaceToken_gitRepositories_project
=== RUN   TestAccDataSecurityNamespaceToken_gitRepositories_repository
=== PAUSE TestAccDataSecurityNamespaceToken_gitRepositories_repository
=== RUN   TestAccDataSecurityNamespaceToken_build_project
=== PAUSE TestAccDataSecurityNamespaceToken_build_project
=== RUN   TestAccDataSecurityNamespaceToken_build_path
=== PAUSE TestAccDataSecurityNamespaceToken_build_path
=== RUN   TestAccDataSecurityNamespaceToken_build_definition
=== PAUSE TestAccDataSecurityNamespaceToken_build_definition
=== RUN   TestAccDataSecurityNamespaceToken_css
=== PAUSE TestAccDataSecurityNamespaceToken_css
=== RUN   TestAccDataSecurityNamespaceToken_iteration
=== PAUSE TestAccDataSecurityNamespaceToken_iteration
=== RUN   TestAccDataSecurityNamespaceToken_tagging_project
=== PAUSE TestAccDataSecurityNamespaceToken_tagging_project
=== RUN   TestAccDataSecurityNamespaceToken_tagging_collection
=== PAUSE TestAccDataSecurityNamespaceToken_tagging_collection
=== RUN   TestAccDataSecurityNamespaceToken_serviceHooks_project
=== PAUSE TestAccDataSecurityNamespaceToken_serviceHooks_project
=== RUN   TestAccDataSecurityNamespaceToken_serviceHooks_collection
=== PAUSE TestAccDataSecurityNamespaceToken_serviceHooks_collection
=== RUN   TestAccDataSecurityNamespaceToken_workItemQueryFolders
=== PAUSE TestAccDataSecurityNamespaceToken_workItemQueryFolders
=== RUN   TestAccDataSecurityNamespaceToken_analytics
=== PAUSE TestAccDataSecurityNamespaceToken_analytics
=== RUN   TestAccDataSecurityNamespaceToken_analyticsViews
=== PAUSE TestAccDataSecurityNamespaceToken_analyticsViews
=== RUN   TestAccDataSecurityNamespaceToken_process
=== PAUSE TestAccDataSecurityNamespaceToken_process
=== RUN   TestAccDataSecurityNamespaceToken_auditLog
=== PAUSE TestAccDataSecurityNamespaceToken_auditLog
=== RUN   TestAccDataSecurityNamespaceToken_buildAdministration
=== PAUSE TestAccDataSecurityNamespaceToken_buildAdministration
=== RUN   TestAccDataSecurityNamespaceToken_server
=== PAUSE TestAccDataSecurityNamespaceToken_server
=== RUN   TestAccDataSecurityNamespaceToken_versionControlPrivileges
=== PAUSE TestAccDataSecurityNamespaceToken_versionControlPrivileges
=== CONT  TestAccDataSecurityNamespaceToken_collection
=== CONT  TestAccDataSecurityNamespaceToken_serviceHooks_project
=== CONT  TestAccDataSecurityNamespaceToken_build_definition
=== CONT  TestAccDataSecurityNamespaceToken_iteration
=== CONT  TestAccDataSecurityNamespaceToken_css
=== CONT  TestAccDataSecurityNamespaceToken_tagging_project
=== CONT  TestAccDataSecurityNamespaceToken_gitRepositories_repository
=== CONT  TestAccDataSecurityNamespaceToken_build_path
=== CONT  TestAccDataSecurityNamespaceToken_tagging_collection
=== CONT  TestAccDataSecurityNamespaceToken_build_project
=== CONT  TestAccDataSecurityNamespaceToken_process
=== CONT  TestAccDataSecurityNamespaceToken_gitRepositories_project
--- PASS: TestAccDataSecurityNamespaceToken_process (2.20s)
=== CONT  TestAccDataSecurityNamespaceToken_workItemQueryFolders
--- PASS: TestAccDataSecurityNamespaceToken_tagging_collection (2.39s)
=== CONT  TestAccDataSecurityNamespaceToken_analyticsViews
--- PASS: TestAccDataSecurityNamespaceToken_collection (2.42s)
=== CONT  TestAccDataSecurityNamespaceToken_analytics
--- PASS: TestAccDataSecurityNamespaceToken_analytics (23.18s)
=== CONT  TestAccDataSecurityNamespaceToken_server
--- PASS: TestAccDataSecurityNamespaceToken_server (1.04s)
=== CONT  TestAccDataSecurityNamespaceToken_versionControlPrivileges
--- PASS: TestAccDataSecurityNamespaceToken_versionControlPrivileges (1.04s)
=== CONT  TestAccDataSecurityNamespaceToken_buildAdministration
--- PASS: TestAccDataSecurityNamespaceToken_buildAdministration (1.24s)
=== CONT  TestAccDataSecurityNamespaceToken_auditLog
--- PASS: TestAccDataSecurityNamespaceToken_auditLog (1.02s)
=== CONT  TestAccDataSecurityNamespaceToken_serviceHooks_collection
--- PASS: TestAccDataSecurityNamespaceToken_serviceHooks_collection (2.26s)
=== CONT  TestAccDataSecurityNamespaceToken_project
--- PASS: TestAccDataSecurityNamespaceToken_tagging_project (32.24s)
--- PASS: TestAccDataSecurityNamespaceToken_build_project (32.51s)
--- PASS: TestAccDataSecurityNamespaceToken_serviceHooks_project (32.67s)
--- PASS: TestAccDataSecurityNamespaceToken_build_path (32.72s)
--- PASS: TestAccDataSecurityNamespaceToken_iteration (32.74s)
--- PASS: TestAccDataSecurityNamespaceToken_gitRepositories_project (32.78s)
--- PASS: TestAccDataSecurityNamespaceToken_css (32.80s)
--- PASS: TestAccDataSecurityNamespaceToken_analyticsViews (32.39s)
--- PASS: TestAccDataSecurityNamespaceToken_workItemQueryFolders (32.64s)
--- PASS: TestAccDataSecurityNamespaceToken_gitRepositories_repository (35.88s)
--- PASS: TestAccDataSecurityNamespaceToken_build_definition (36.76s)
--- PASS: TestAccDataSecurityNamespaceToken_project (22.42s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	54.633s
```

I have two minor questions want to confirm with you before I merge this PR:

1. What is your source (document) of these namespace tokens? What I've found is https://learn.microsoft.com/en-us/azure/devops/organizations/security/namespace-reference?view=azure-devops, which though contains a comprehensive list of namespaces, while the description of each namespace's token is inconsistent and incomplete, e.g. the `Build` doesn't mention `path`.
2. Since this data source is meant to be a generic data source, which expects to cover *all* possible namespaces and token formats. The current implementation seems just combines the existing specific token data sources into one. Instead, maybe users shall be able to figure out the token format themselves and specify the token directly?

aklakina:
As for the first question:
I'm building a full IaC repository for managing azure devops and thus I had to implement some custom resources (some of which are not even published yet as a pr). Because of this, most of my prs are built on observations from the api requests going to the service. The tokens here are also an observation about how the service handles these tokens.

As for your second question:
This is just a static data source which is not even prerequisite for my next pr. This token will be consumed by the 3rd pr in this group and is handled as a plain string. So if a user wants a custom token to be used, there they can specify it.

magodo (approved):
LGTM, thanks!

PR #1470, opened 2025-12-19
New resource: `azuredevops_workitemtrackingprocess_process_permissions`

Fresa:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [ ] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## What about the current behavior has changed?
Manage [ACL permissions for processes](https://learn.microsoft.com/en-us/azure/devops/organizations/security/set-permissions-access-work-tracking?view=azure-devops#customize-an-inherited-process).

Issue Number: [1446](https://github.com/microsoft/terraform-provider-azuredevops/issues/1446)

## Does this introduce a change to `go.mod`, `go.sum` or `vendor/`?

- [ ] Yes
- [x] No

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Any relevant logs, error output, etc?
N/A

## Other information
Other unrelated unit tests are failing on main.

Issue #354, opened 2021-04-05
Add support for managing release pipeline folders and permissions


ajhall:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I'd like to use Terraform to manage release pipeline folders. We often create an empty release folder, then assign permissions to that folder so a specific team can manage its contents.

I'd also like to be able to set permissions on those folders, similar to the existing `azuredevops_build_definition_permissions` resource, but applicable to release pipelines and folders.

### New or Affected Resource(s)

- azuredevops_release_definition_folder (new)
- azuredevops_release_definition_permissions (new)

### Potential Terraform Configuration

```hcl
resource "azuredevops_release_definition_folder" "folder" {
  project_id = azuredevops_project.project.id
  path       = "\\ExampleFolder"
}

resource "azuredevops_release_definition_permissions" "folder" {
  project_id  = azuredevops_project.project.id
  principal   = data.azuredevops_group.project-readers.id

  path        = azuredevops_release_definition_folder.folder

  permissions = {
    ViewReleaseDefinition        = "Allow"
    EditReleaseDefinition        = "Allow"
    DeleteReleaseDefinition      = "Allow"
    ManageReleaseApprovers       = "Allow"
    ManageReleases               = "Allow"
    ViewReleases                 = "Allow"
    CreateReleases               = "Allow"
    EditReleaseEnvironment       = "Allow"
    DeleteReleaseEnvironment     = "Allow"
    AdministerReleasePermissions = "Allow"
    DeleteReleases               = "Allow"
    ManageDeployments            = "Allow"
    ManageReleaseSettings        = "Allow"
    ManageTaskHubExtension       = "Allow"
  }
}
```

### References

- https://github.com/microsoft/terraform-provider-azuredevops/issues/49
- https://docs.microsoft.com/en-us/rest/api/azure/devops/release/folders?view=azure-devops-rest-6.0
- https://docs.microsoft.com/en-us/rest/api/azure/devops/security/access%20control%20lists?view=azure-devops-rest-6.0
[...]

Issue #340, opened 2021-03-24
Support setting all group permissions


mintxelas:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently only some of the permissions listed in the Group's Project Permission tab can be configured during deployment.
"Create tag definition" and the "Analytics" permissions cannot be set.
Also, the documentation is missing the list of possible values (https://www.pulumi.com/docs/reference/pkg/azuredevops/projectpermissions/).

### New or Affected Resource(s)

https://www.pulumi.com/docs/reference/pkg/azuredevops/projectpermissions/ should contain the list of all the group permissions

* azuredevops_XXXXX

### Potential Terraform Configuration

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### References

https://www.pulumi.com/docs/reference/pkg/azuredevops/projectpermissions/
https://docs.microsoft.com/en-us/azure/devops/organizations/security/namespace-reference?view=azure-devops#project-level-namespaces-and-permissions

* #0000

drdamour:
also it's unclear to me what these settings are for and where they show up in the web ux

    # i THINK These are piplines things, accessible within the pipline (though it doesn't match up completely)
    # Administer a build
    # ADMINISTER_BUILD	
    # Start a build
    # NOTE: this might be Queue Build
    # START_BUILD	
    # Edit build quality
    # EDIT_BUILD_STATUS	
    # Write to build operational store
    # NOTE: this might be Update Buil Infomration
    # UPDATE_BUILD	

    # These i have no idea where they are set
    # MANAGE_SYSTEM_PROPERTIES	Manage system project properties
    # BYPASS_PROPERTY_CACHE	Bypass project property cache
    # AGILETOOLS_BACKLOG	Agile backlog management.
    # AGILETOOLS_PLANS	Agile plans.

mintxelas:
This is the most complete list I found: https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/project_permissions

PR #1473, opened 2025-12-22
Add new resource: `azuredevops_security_permissions` for allowing low-level permission management

aklakina:
## All Submissions:

This is PR#3 of 3 PRs introducing generic security management capabilities for Azure DevOps, providing low-level access to the security namespace system. (see #1449 for parent pr)

**Resource:**
- `azuredevops_security_permissions` - Generic permissions management across any security namespace, with validation of permission names and support for both replace and merge strategies

This resource exposes the [security permissions](https://learn.microsoft.com/en-us/rest/api/azure/devops/security/permissions) api endpoint. Based on the documentation one can not assign permissions, but the sdk do offer this capability.

Issue Number:
- #340
- #354
- #404
- #521
- #850
- #998
- #1023
- #1177
- #1217
- #1344
- #1470 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


aklakina:
Hi @magodo,

Can you have a look at this pr when you have the time?

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+~> **Note** This is a low-level generic permissions resource. For specific resource types, consider using the dedicated permission resources such as `azuredevops_project_permissions`, `azuredevops_git_permissions`, `azuredevops_build_definition_permissions`, etc.
+
+~> **Note** Permissions can be assigned to group principals and not to single user principals.
+
+## Example Usage
+
+### Collection-level Permissions
+
+```hcl
+data "azuredevops_security_namespaces" "all" {
```
magodo: Can we use the `azuredevops_security_namespace` data source instead?

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+data "azuredevops_group" "example_readers" {
+  project_id = data.azuredevops_project.example.id
+  name       = "Readers"
+}
+
+resource "azuredevops_security_permissions" "project_perms" {
+  namespace_id = "52d39943-cb85-4d7f-8fa8-c6baac873819" # Project namespace
```
magodo: Could this be queried from `azuredevops_security_namespace`?

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+data "azuredevops_group" "example_contributors" {
+  project_id = data.azuredevops_project.example.id
+  name       = "Contributors"
+}
+
+resource "azuredevops_security_permissions" "git_perms" {
+  namespace_id = "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87" # Git Repositories namespace
```
magodo: Same here

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `namespace_id` - (Required) The ID of the security namespace. Use the `azuredevops_security_namespaces` data source to discover available namespaces. Changing this forces a new resource to be created.
+
+* `token` - (Required) The security token for the resource. Use the `azuredevops_security_namespace_token` data source to generate tokens for specific resources. Changing this forces a new resource to be created.
+
+* `principal` - (Required) The descriptor or identity ID of the principal (user or group). Changing this forces a new resource to be created.
```
magodo, line 170: You have called out in the begining of this file:

> **Note** Permissions can be assigned to group principals and not to single user principals.

While here you were indicating a user is also a valid principal. Could you please confirm?

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+
+* `permissions` - (Required) A map of permission names to permission values. All permission names specified must be valid for the given namespace, or an error will be returned. Permission values must be one of:
+  - `Allow` (or `allow`, `ALLOW`) - Grant the permission
+  - `Deny` (or `deny`, `DENY`) - Explicitly deny the permission
+  - `NotSet` (or `notset`, `NOTSET`) - Remove the permission (inherit from parent)
```
magodo, line 175: Can we make the value a boolean:

- `true` -> allow
- `false` -> deny
- absent -> notset

website/docs/r/security_permissions.html.markdown
```diff
@@ -0,0 +1,255 @@
...
+
+* `replace` - (Optional) Replace (`true`) or merge (`false`) the permissions with existing permissions. When `true`, all existing permissions for the principal on this token will be replaced with the specified permissions. When `false`, the specified permissions will be merged with existing permissions. Default: `false`.
```
magodo: Can we simply stick on the `replace` mode, which makes more sense for an IaC tool like terraform?

azuredevops/internal/service/security/resource_security_permissions.go
```diff
@@ -0,0 +1,581 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("removing managed permissions: %v", err)
+	}
+
+	log.Printf("[INFO] Successfully removed managed permissions from ACE for principal %s", principal)
+	return nil
+}
+
+// buildActionMap creates a mapping from permission names (and DisplayNames) to their bit values.
```
magodo: This seems over complicated. Instead shall we stick on using `name` as the permission key instead of accepting both the name and display name?

azuredevops/internal/service/security/resource_security_permissions.go
```diff
@@ -0,0 +1,581 @@
...
+		}
+	}
+
+	if err := d.Set("permissions", currentPermissions); err != nil {
+		return fmt.Errorf("setting permissions in state: %v", err)
+	}
+	return nil
+}
+
+func resourceGenericPermissionsDelete(d *schema.ResourceData, m interface{}) error {
```
magodo, line 350: The delete logic is a bit confusing as it seems simply revert the changes to the specified ACE. Say the user enables a previous enabled action, then during delete it will be cleared and reset to be "NotSet", while it was initially "Allowed".

azuredevops/internal/service/security/resource_security_permissions.go
```diff
@@ -0,0 +1,581 @@
...
+		Token                *string                        `json:"token,omitempty"`
+		Merge                *bool                          `json:"merge,omitempty"`
+		AccessControlEntries *[]security.AccessControlEntry `json:"accessControlEntries,omitempty"`
+	}{
+		Token:                &token,
+		Merge:                &bMerge,
+		AccessControlEntries: &[]security.AccessControlEntry{updatedACE},
+	}
+
+	_, err = clients.SecurityClient.SetAccessControlEntries(clients.Ctx, security.SetAccessControlEntriesArgs{
```
magodo, line 475: There requires a retry polling to accomodate the eventual consistency issue.

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/security/resource_security_permissions.go
aklakina: That was my original approach. The problem with that was then the process to map permission settings from ui to code involves requesting the namespaces api. From there you would need to find the name of the permission bit you want to set. User experience is much better like this.

azuredevops/internal/service/security/resource_security_permissions.go
aklakina: I don't fully understand this comment. Do mean that it _should_ act like this, or do you mean that it _looks like_ it acts like this?

If it _should_ then I don't see the reason why. If a resource is managed by terraform and is no longer part of a configuration, I think the remote state should match this independently of what was the original state. Of course this resource could be a bit of an exception here, because one does not actively import this resource so there is no such a thing as "already exists".

website/docs/r/security_permissions.html.markdown
aklakina: This would not allow explicitly setting a permission to notset which may be useful for inheritance.

website/docs/r/security_permissions.html.markdown
aklakina: I have taken this implementation detail from the group_membership and team_membership resource. Of course I could modify it to default: true, or remove it completely and just keep the true state. Which one do you prefer?

website/docs/r/security_permissions.html.markdown
aklakina: That was a faulty statement, it supports any principal object.

website/docs/r/security_permissions.html.markdown
magodo: Yes, let's make it defaults to true.

azuredevops/internal/service/security/resource_security_permissions.go
magodo: If we have a decent document for users to reference, there will be no need for them to map the display name from UI to code?

azuredevops/internal/service/security/resource_security_permissions.go
magodo: I meant it looks like it which seems incorrect.

azuredevops/internal/service/security/resource_security_permissions.go
aklakina: Not entirely because for example, the namespace api returns
```json
  {
      "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87",
      "name": "Git Repositories",
      "displayName": "Git Repositories",
      "separatorValue": "/",
      "elementLength": -1,
      "writePermission": 8192,
      "readPermission": 2,
      "dataspaceCategory": "Git",
      "actions": [
        {
          "bit": 1,
          "name": "Administer",
          "displayName": "Administer",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 2,
          "name": "GenericRead",
          "displayName": "Read",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 4,
          "name": "GenericContribute",
          "displayName": "Contribute",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 8,
          "name": "ForcePush",
          "displayName": "Force push (rewrite history, delete branches and tags)",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 16,
          "name": "CreateBranch",
          "displayName": "Create branch",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 32,
          "name": "CreateTag",
          "displayName": "Create tag",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 64,
          "name": "ManageNote",
          "displayName": "Manage notes",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 128,
          "name": "PolicyExempt",
          "displayName": "Bypass policies when pushing",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 256,
          "name": "CreateRepository",
          "displayName": "Create repository",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 512,
          "name": "DeleteRepository",
          "displayName": "Delete or disable repository",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 1024,
          "name": "RenameRepository",
          "displayName": "Rename repository",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 2048,
          "name": "EditPolicies",
          "displayName": "Edit policies",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 4096,
          "name": "RemoveOthersLocks",
          "displayName": "Remove others' locks",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 8192,
          "name": "ManagePermissions",
          "displayName": "Manage permissions",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 16384,
          "name": "PullRequestContribute",
          "displayName": "Contribute to pull requests",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 32768,
          "name": "PullRequestBypassPolicy",
          "displayName": "Bypass policies when completing pull requests",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 65536,
          "name": "ViewAdvSecAlerts",
          "displayName": "Advanced Security: view alerts",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 131072,
          "name": "DismissAdvSecAlerts",
          "displayName": "Advanced Security: manage and dismiss alerts",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        },
        {
          "bit": 262144,
          "name": "ManageAdvSecScanning",
          "displayName": "Advanced Security: manage settings",
          "namespaceId": "2e9eb7ed-3c0a-47d4-87c1-0ffdd275fd87"
        }
      ],
      "structureValue": 1,
      "extensionType": "Microsoft.TeamFoundation.Git.Server.Plugins.GitSecurityNamespaceExtension",
      "isRemotable": true,
      "useTokenTranslator": true,
      "systemBitMask": 0
    }
```
But this does not specify which permissions are applicable on the project, repository or branch/tag scope. For example administer does nothing on repository scope (but does not fail either when you try to set it). Instead Users can use the display name on the ui for the specific permission bit they would like to set. This way it is more evident what they want to set. Proper documentation would solve this issue, true but this makes it easier for the users to use the resource too. Also there is a failsafe that explicitly warns/errors the user if a displayname matches for 2 different names.

azuredevops/internal/service/security/resource_security_permissions.go
aklakina: The algorithm queries the already set bits and deletes the one present in the configuration. I think this is how it should work since if you set the replace mode to false, it should keep the originally set bits while deleting the ones defined.

azuredevops/internal/service/security/resource_security_permissions.go
magodo: I want to avoid this alternatives which might seem to be flexible, whilst causing confusion and maintainance burden.

azuredevops/internal/service/security/resource_security_permissions.go
magodo: This requires the user to be extream careful to not specify any already enabled permission in the TF config if the replace mode is set to false. That's why I still prefer to stick on `replace` mode only.

aklakina: So then should I delete the `replace = false` option?
Or is it okay if I explicitly extend the documentation with a note saying `replace = false` only deletes permissions that are present in the config?

aklakina:
hi @magodo 

I don't think the build validation errors are coming from this pr. Can you confirm it?

azuredevops/internal/service/security/resource_security_permissions.go
aklakina: Removed this feature, just stick with name property of the security mapping.

azuredevops/internal/service/security/resource_security_permissions.go
magodo: That sounds reasonable.

magodo:
No, it is not related to your PR and is solved in https://github.com/microsoft/terraform-provider-azuredevops/pull/1510. Please merge the `main` branch.

magodo:
@aklakina Now th CI failed because you have imported the legacy test package in your file: ./internal/acceptancetests/resource_security_permissions_test.go.

Please replace `github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource` to `github.com/hashicorp/terraform-plugin-testing/helper/resource`.

magodo:
The acceptance tests have issues with the configuration, can you ensure they can pass?

aklakina:
Hi @magodo,

Sorry, I've had a lot to do last week. I've fixed the acceptance tests and ran it against our cloud instance:
image

magodo (approved):
LGTM!

```shell
terraform-provider-azuredevops on  feature/r/security-permissions via 🐹 v1.26.0-X:nodwarf5 took 11s
💢 TF_ACC=1 go test -v -run='TestAccSecurityPermissions_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccSecurityPermissions_ProjectNamespace
=== PAUSE TestAccSecurityPermissions_ProjectNamespace
=== RUN   TestAccSecurityPermissions_ProjectNamespaceUpdate
=== PAUSE TestAccSecurityPermissions_ProjectNamespaceUpdate
=== RUN   TestAccSecurityPermissions_GitRepositoryNamespace
=== PAUSE TestAccSecurityPermissions_GitRepositoryNamespace
=== RUN   TestAccSecurityPermissions_WithReplace
=== PAUSE TestAccSecurityPermissions_WithReplace
=== RUN   TestAccSecurityPermissions_BuildNamespace
=== PAUSE TestAccSecurityPermissions_BuildNamespace
=== RUN   TestAccSecurityPermissions_MultiplePermissionSets
=== PAUSE TestAccSecurityPermissions_MultiplePermissionSets
=== CONT  TestAccSecurityPermissions_ProjectNamespace
=== CONT  TestAccSecurityPermissions_GitRepositoryNamespace
=== CONT  TestAccSecurityPermissions_WithReplace
=== CONT  TestAccSecurityPermissions_MultiplePermissionSets
=== CONT  TestAccSecurityPermissions_BuildNamespace
=== CONT  TestAccSecurityPermissions_ProjectNamespaceUpdate
--- PASS: TestAccSecurityPermissions_MultiplePermissionSets (29.64s)
--- PASS: TestAccSecurityPermissions_BuildNamespace (29.64s)
--- PASS: TestAccSecurityPermissions_ProjectNamespace (29.66s)
--- PASS: TestAccSecurityPermissions_GitRepositoryNamespace (31.00s)
--- PASS: TestAccSecurityPermissions_ProjectNamespaceUpdate (35.66s)
--- PASS: TestAccSecurityPermissions_WithReplace (35.68s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        35.688s
```

PR #1474, opened 2025-12-23
New resource: `azuredevops_workitemtrackingprocess_control`

Fresa:
## All Submissions:

Add [controls](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/customize-process-field?view=azure-devops) (including [custom controls](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/custom-controls-process?view=azure-devops)) to work item types.

~~This PR is stacked upon [1462](https://github.com/microsoft/terraform-provider-azuredevops/pull/1462) which probably should be merged first.~~

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


magodo:
@Fresa Thank you for this contribution! Do you mind to split this PR into a few small PRs where each one contain only a single resource, to make it easier to track/review?

Fresa:
@magodo The PR is stacked on [1462](https://github.com/microsoft/terraform-provider-azuredevops/pull/1462), so if you could review that first! 🎁

Fresa:
@magodo I've updated the PR with the latest from upstream/main, it should be good for review now!

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_control_test.go
```diff
@@ -0,0 +1,334 @@
+//go:build (all || resource_workitemtrackingprocess || resource_workitemtrackingprocess_control) && !exclude_resource_workitemtrackingprocess
+// +build all resource_workitemtrackingprocess resource_workitemtrackingprocess_control
+// +build !exclude_resource_workitemtrackingprocess
+
```
magodo: These are not necessary.

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
```diff
@@ -0,0 +1,481 @@
...
+		Id:       converter.String(d.Get("control_id").(string)),
+		Visible:  converter.Bool(d.Get("visible").(bool)),
+		ReadOnly: converter.Bool(d.Get("read_only").(bool)),
+	}
+
+	if v, ok := d.GetOk("label"); ok {
+		control.Label = converter.String(v.(string))
+	}
+	//nolint:staticcheck // SA1019: d.GetOkExists is deprecated but required to distinguish between unset and zero value
+	if v, ok := d.GetOkExists("order"); ok {
```
magodo: Instead of doing this, you shall use the `d.GetRawConfig().AsValueMap()`. See an example at: https://github.com/microsoft/terraform-provider-azuredevops/blob/56c96ad5d3b5f7581b6e902eafd652657a01e553/azuredevops/internal/service/core/resource_project_pipeline_settings.go#L128-L132

azuredevops/internal/service/workitemtrackingprocess/retry.go
```diff
@@ -0,0 +1,52 @@
...
+func retryOnNotFound(ctx context.Context, timeout time.Duration, f func() error) error {
+	return retryOnCondition(ctx, timeout, f, func(err error) bool {
+		return utils.ResponseWasStatusCode(err, http.StatusNotFound)
+	})
+}
+
+// retryOnContributionNotFound retries the given function when Azure DevOps returns
+// a VS403120 error ("Contribution does not exist"). This handles eventual consistency
+// where a newly installed extension's contributions may not be immediately available.
+func retryOnContributionNotFound(ctx context.Context, timeout time.Duration, f func() error) error {
```
magodo, line 48: Is this necessary as the extension is not automated in Terraform?

azuredevops/internal/service/workitemtrackingprocess/retry.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// retryOnNotFound retries the given function when Azure DevOps returns
+// a 404 Not Found status code. This handles eventual consistency where
+// a newly created resource may not immediately be visible.
+func retryOnNotFound(ctx context.Context, timeout time.Duration, f func() error) error {
+	return retryOnCondition(ctx, timeout, f, func(err error) bool {
+		return utils.ResponseWasStatusCode(err, http.StatusNotFound)
+	})
+}
```
magodo, line 43: These twos indicate the implementation of the depending resource returns without waiting until it is fully functional. Instead of adding a retry here, can we root cause the issue at the depending resource instead?

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
```diff
@@ -0,0 +1,481 @@
...
+
+	args := workitemtrackingprocess.CreateControlInGroupArgs{
+		ProcessId:  converter.UUID(d.Get("process_id").(string)),
+		WitRefName: converter.String(d.Get("work_item_type_reference_name").(string)),
+		GroupId:    converter.String(d.Get("group_id").(string)),
+		Control:    &control,
+	}
+
+	var createdControl *workitemtrackingprocess.Control
+	err := retryOnContributionNotFound(ctx, d.Timeout(schema.TimeoutCreate), func() error {
```
magodo, line 211: Instead of using `schema.TimeoutCeate`, we shall use the timeout deduced from the `ctx`.

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+
+	if createdControl.Id == nil {
+		return diag.Errorf(" Created control has no ID")
+	}
+
+	d.SetId(*createdControl.Id)
+
+	// Retry reading the control to handle eventual consistency
+	err = retryOnNotFound(ctx, d.Timeout(schema.TimeoutCreate), func() error {
```
magodo, line 227: Instead of retry on reading and flattening the control resource, we shall use the `retry.StateChangeConf` to refresh the state of the control until it is accessible, then call the read() at the end.

Though they are technically doing the same thing, but just want to make the code consistent across the code base.

Example: https://github.com/microsoft/terraform-provider-azuredevops/blob/56c96ad5d3b5f7581b6e902eafd652657a01e553/azuredevops/internal/service/serviceendpoint/commons.go#L82-L90

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
```diff
@@ -0,0 +1,481 @@
...
+		return readResourceControlWithError(ctx, d, m)
+	})
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
```
magodo, line 233: Then this shall be changed to:


```suggestion
	return readResourceControl(ctx, d, m)
```

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
```diff
@@ -75,6 +75,111 @@ func ResourceGroup() *schema.Resource {
...
+						},
+						"inherited": {
+							Type:        schema.TypeBool,
+							Computed:    true,
+							Description: "A value indicating whether this control has been inherited from a parent layout.",
+						},
+						"overridden": {
+							Type:        schema.TypeBool,
+							Computed:    true,
+							Description: "A value indicating whether this control has been overridden by a child layout.",
```
magodo, line 141: Wondering why these twos are not Optional here as opposed to the standalone control resource?

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
```diff
@@ -75,6 +75,111 @@ func ResourceGroup() *schema.Resource {
...
+						},
+						"read_only": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     false,
+							Description: "A value indicating if the control is read only.",
+						},
+						"order": {
+							Type:        schema.TypeInt,
+							Computed:    true,
```
magodo, line 120: Wondering why this isn't optional?

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_control_test.go
Fresa: removed

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
Fresa: cool, done!

azuredevops/internal/service/workitemtrackingprocess/retry.go
Fresa: Not sure what you mean by that, but this is necessary due to eventual consistency between the WIT control api and the extension management api. You can verify this behavior using [this test](https://github.com/microsoft/terraform-provider-azuredevops/blob/98ec16f9386965f4c31db86ebe022302e0dd9e46/azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_control_test.go#L164) by replacing VS403120 with some other string. It will eventually throw something like  `Error:  Creating control. Error VS403120: Contribution "ms-devlabs.vsts-extensions-multivalue-control.multivalue-form-control" does not exist.`

azuredevops/internal/service/workitemtrackingprocess/retry.go
Fresa: The root cause is eventual consistency as far as I can tell. The first one is a bit cryptic, but as the comment states, this seems to happen from time to time due to resource dependencies. Searching for this error message you'll find a couple of reported issues regarding this error, but no definite solutions. It's not documented anywhere as far as I can tell. 

You can provoke this error with [this test](https://github.com/microsoft/terraform-provider-azuredevops/blob/98ec16f9386965f4c31db86ebe022302e0dd9e46/azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_control_test.go#L102) when removing [this call](https://github.com/microsoft/terraform-provider-azuredevops/blob/98ec16f9386965f4c31db86ebe022302e0dd9e46/azuredevops/internal/service/workitemtrackingprocess/resource_group.go#L418) or [this call](https://github.com/microsoft/terraform-provider-azuredevops/blob/98ec16f9386965f4c31db86ebe022302e0dd9e46/azuredevops/internal/service/workitemtrackingprocess/resource_control.go#L396).
`Error:  Delete group Test Group. Error TF401349: An unexpected error has occurred, please verify your request and try again.`
`Error:  Delete control. Error TF401349: An unexpected error has occurred, please verify your request and try again.`

If you look at the resources managed by the test, you see that they all have implicit references, so they should be destroyed in the expected order.

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
Fresa: this method is calling into the SDK's [retry helper](https://github.com/microsoft/terraform-provider-azuredevops/blob/98ec16f9386965f4c31db86ebe022302e0dd9e46/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry/wait.go#L18) which requires both `ctx` (with cancellation) and a timeout duration. Is it possible to ignore the timeout duration, or can it be inferred by the cancellation token?

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
Fresa: shouldn't `RetryContext` be the preferred convention, considering it is simpler to use and understand? The example you posted both uses StateChangeConf and RetryContext ;) The change entails 15 extra lines of code, without really any benefits IMO.

I'll commit the change of course if you insist!

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
Fresa: The order is determined based on the order of the controls in the list. The description: `Order in which the control appears in the group. Determined by the position in the control list.`


Realized I forgot to sort the returned list though, so I've added that. Also discovered another eventual consistency issue regarding said list that I also fixed, even though one of the acceptance tests might be flakey because of it running refresh immediately after create/update. I added a note about it.

azuredevops/internal/service/workitemtrackingprocess/resource_group.go
Fresa: nice catch! I realized when testing groups that these two does nothing, hence why they were defined as computed only in the group schema. I've removed the optional directive in the control resource. I tried changing one of the acceptance tests for controls to verify that it behaved the same (ignores them), and it does. One could argue that we could remove them all together, I think these are only applicable for controls inherited from the inherited WIT.

azuredevops/internal/service/workitemtrackingprocess/retry.go
magodo: I've noticed that this is a systematic issue, where a second call (even with no delay) will succeed. I doubt there is some internal issue at the service side instead of an eventual consistency issue.

azuredevops/internal/service/workitemtrackingprocess/retry.go
magodo: Confirm the first one is intermittently happening.

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
magodo: Seems what you were doing here matches the document's suggestion: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#retry. Though we can use another timeout value, or deduce it from `ctx.Deadline()`.

azuredevops/internal/service/workitemtrackingprocess/resource_control.go
magodo: Let's keep the retry function.

azuredevops/internal/service/workitemtrackingprocess/retry.go
magodo: Confirm the second one is intermittently happening for `TestAccWorkitemtrackingprocessControl_Contribution`, so assuming it is also related to the extension.

magodo (approved):

magodo:
The control test passes:

```shell
terraform-provider-azuredevops on  add-controllers via 🐹 v1.25.5 took 12s
💢 TF_ACC=1 go test -v -run='TestAccWorkitemtrackingprocessControl_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkitemtrackingprocessControl_Basic
=== PAUSE TestAccWorkitemtrackingprocessControl_Basic
=== RUN   TestAccWorkitemtrackingprocessControl_Update
=== PAUSE TestAccWorkitemtrackingprocessControl_Update
=== RUN   TestAccWorkitemtrackingprocessControl_Move
=== PAUSE TestAccWorkitemtrackingprocessControl_Move
=== RUN   TestAccWorkitemtrackingprocessControl_Contribution
=== PAUSE TestAccWorkitemtrackingprocessControl_Contribution
=== CONT  TestAccWorkitemtrackingprocessControl_Basic
=== CONT  TestAccWorkitemtrackingprocessControl_Update
=== CONT  TestAccWorkitemtrackingprocessControl_Move
=== CONT  TestAccWorkitemtrackingprocessControl_Contribution
--- PASS: TestAccWorkitemtrackingprocessControl_Basic (2.41s)
--- PASS: TestAccWorkitemtrackingprocessControl_Move (5.26s)
--- PASS: TestAccWorkitemtrackingprocessControl_Update (5.64s)
--- PASS: TestAccWorkitemtrackingprocessControl_Contribution (7.16s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	7.164s
```

While the group test case failed:

```shell
💤 TF_ACC=1 go test -v -run='TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
=== PAUSE TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
=== CONT  TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
    resource_workitemtrackingprocess_group_test.go:167: Step 1/2 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout


        Terraform used the selected providers to generate the following execution
        plan. Resource actions are indicated with the following symbols:
          ~ update in-place

          # azuredevops_workitemtrackingprocess_group.test will be updated in-place
          ~ resource "azuredevops_workitemtrackingprocess_group" "test" {
                id                            = "0ed3a6ae-e0ac-481e-832a-15c7a992b6a6"
                # (7 unchanged attributes hidden)

              + control {
                  + id              = "ms-devlabs.vsts-extensions-multivalue-control.multivalue-form-control"
                  + is_contribution = true
                  + read_only       = false
                  + visible         = true

                  + contribution {
                      + contribution_id           = "ms-devlabs.vsts-extensions-multivalue-control.multivalue-form-control"
                      + height                    = 50
                      + inputs                    = {
                          + "FieldName" = "System.Tags"
                          + "Values"    = "Option1;Option2;Option3"
                        }
                      + show_on_deleted_work_item = false
                    }
                }

                # (5 unchanged blocks hidden)
            }

        Plan: 0 to add, 1 to change, 0 to destroy.
--- FAIL: TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes (4.00s)
FAIL
FAIL	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	4.008s
FAIL
```

This can be solved by using the [StateChangeConf](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts#statechangeconf), with a reasonale `ContinuousTargetOccurence`.

Fresa:
> ContinuousTargetOccurence

@magodo I've added ContinuousTargetOccurence: 4 when reading a group after create that defines controls inline, anything less still triggers the inconsistency intermittently. I'm not convinced 4 won't fail at some time either, but we can't retry indefinitely, and 4 already adds quite a lot of latency... I avoided ContinuousTargetOccurence intentionally before because I think this will only materialize in the acceptance tests. Users will never refresh resources that fast. Let me know what you think!

```
=== RUN   TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
=== PAUSE TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
=== CONT  TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes
[DEBUG] Expected control count: 6, Actual control count: 6
[DEBUG] Expected control count: 6, Actual control count: 6
[DEBUG] Expected control count: 6, Actual control count: 6
[DEBUG] Expected control count: 6, Actual control count: 5
[DEBUG] Expected control count: 6, Actual control count: 6
--- PASS: TestAccWorkitemtrackingprocessGroup_WithMultipleControlTypes (10.62s)
```

Fresa:
@magodo btw when you merge this, could you also create a release?

magodo (approved):

magodo:
@Fresa I'll create a release after fixing some other issues.

Issue #839, opened 2023-07-21
Implement work item custom_fields


ghs:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Custom fields in work items are of specific shape which as far I can see and read through the code is not implemented: https://github.com/microsoft/terraform-provider-azuredevops/pull/659/files#diff-d52cafbb405887e133170ccdec83c29154d37e34ac7b4e47e95b1942e93f3779.

The existing documentation also doesn't give a lot of info other than `custom_fields = {
    example : "example"
  }` for proper definition of them.

From ADO, custom fields, when created, could contain the following properties: is it from existing field or a new one, **name**, **description**, **type**, **required** with default value, and layout with multiple properties.

### New or Affected Resource(s)

azuredevops_workitem

### References

https://github.com/microsoft/terraform-provider-azuredevops/pull/659

xuzhang3:
potential config:
```hcl
resource azuredevops_workitem_custom_field {

}
```
API: https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/fields/add?view=azure-devops-rest-7.0&tabs=HTTP

Fresa:
I'm going to take a stab at this.

Fields are added to an organization via [these APIs](https://learn.microsoft.com/en-us/rest/api/azure/devops/wit/fields?view=azure-devops-rest-7.1).

The API linked to adds fields (created in the organization) to a work item type of an inherited process. Inherited processes inherits (system) fields from the system process it inherits from by default. I'm probably going to implement that too later.

Related: [1446](https://github.com/microsoft/terraform-provider-azuredevops/issues/1446)

PR #1475, opened 2025-12-29
New resource: `azuredevops_workitemtracking_field`

Fresa:
## All Submissions:

Adds support for managing [custom fields](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/add-custom-field?view=azure-devops).

Issue Number: [839](https://github.com/microsoft/terraform-provider-azuredevops/issues/839)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


Fresa:
@magodo this one is ready for review, when you have time!

azuredevops/internal/acceptancetests/testutils/field.go
```diff
@@ -0,0 +1,56 @@
+package testutils
```
magodo: Can we move these to the resource test file instead of in the `testutils` package?

azuredevops/internal/acceptancetests/testutils/field.go
```diff
@@ -0,0 +1,56 @@
...
+	timeout := 10 * time.Second
+
+	for _, resource := range s.RootModule().Resources {
+		if resource.Type != "azuredevops_workitemtracking_field" {
+			continue
+		}
+
+		referenceName := resource.Primary.ID
+
+		err := retry.RetryContext(clients.Ctx, timeout, func() *retry.RetryError {
```
magodo: Can we move this retry to the Delete function of the resource as a post delete polling, instead?

azuredevops/internal/acceptancetests/resource_workitemtracking_field_test.go
```diff
@@ -0,0 +1,250 @@
...
+					resource.TestCheckNoResourceAttr(tfNode, "description"),
+					resource.TestCheckResourceAttr(tfNode, "usage", "workItem"),
+					resource.TestCheckResourceAttr(tfNode, "read_only", "false"),
+					resource.TestCheckResourceAttr(tfNode, "can_sort_by", "true"),
+					resource.TestCheckResourceAttr(tfNode, "is_queryable", "true"),
+					resource.TestCheckResourceAttr(tfNode, "is_identity", "false"),
+					resource.TestCheckResourceAttr(tfNode, "is_picklist", "false"),
+					resource.TestCheckResourceAttr(tfNode, "is_picklist_suggested", "false"),
+					resource.TestCheckNoResourceAttr(tfNode, "picklist_id"),
+					resource.TestCheckResourceAttr(tfNode, "is_locked", "false"),
```
magodo: These non-compute-only attributes will be checked during the import step below, so no need to check them manually here (and in the other steps).

website/docs/r/workitemtracking_field.html.markdown
```diff
@@ -0,0 +1,141 @@
...
+### HTML Field
+
+```hcl
+resource "azuredevops_workitemtracking_field" "notes" {
+  name           = "Custom Notes"
+  reference_name = "Custom.Notes"
+  type           = "html"
+  description    = "Additional notes in HTML format"
+}
+```
```
magodo: These examples seem redundent/unnecessary.

azuredevops/internal/service/workitemtracking/resource_field.go
```diff
@@ -0,0 +1,366 @@
...
+			Payload: &workitemtracking.FieldUpdate{
+				IsDeleted: &isDeleted,
+			},
+		}
+
+		restoredField, err := clients.WorkItemTrackingClient.UpdateWorkItemField(clients.Ctx, args)
+		if err != nil {
+			return diag.Errorf("restoring field %s: %+v", *field.ReferenceName, err)
+		}
+		d.SetId(*restoredField.ReferenceName)
```
magodo, line 228: Can we nil check both `restoreField` and `restoreField.ReferenceName`?

azuredevops/internal/service/workitemtracking/resource_field.go
```diff
@@ -0,0 +1,366 @@
...
+		args := workitemtracking.CreateWorkItemFieldArgs{
+			WorkItemField: field,
+		}
+
+		createdField, err := clients.WorkItemTrackingClient.CreateWorkItemField(clients.Ctx, args)
+		if err != nil {
+			return diag.Errorf("creating field: %+v", err)
+		}
+
+		if createdField.ReferenceName == nil {
```
magodo, line 242: Can we also nil check `createdField`?

azuredevops/internal/service/workitemtracking/resource_field.go
```diff
@@ -0,0 +1,366 @@
...
+
+	field, err := clients.WorkItemTrackingClient.GetWorkItemField(clients.Ctx, args)
+	if err != nil {
+		if utils.ResponseWasNotFound(err) {
+			d.SetId("")
+			return nil
+		}
+		return diag.Errorf("reading field %s: %+v", referenceName, err)
+	}
+
```
magodo, line 270: Can we nil check `field`?

azuredevops/internal/service/workitemtracking/resource_field.go
```diff
@@ -0,0 +1,366 @@
...
+				Optional:    true,
+				Default:     false,
+				Description: "Indicates whether this field is locked for editing. Default: `false`.",
+			},
+			"url": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The URL of the field resource.",
+			},
+			"restore": {
```
magodo, line 148: I doubt the necessity to introduce this field, as restoring a deleted attribute seems to be effectively the same as creating it?

website/docs/r/workitemtracking_field.html.markdown
```diff
@@ -0,0 +1,141 @@
...
+page_title: "AzureDevops: azuredevops_workitemtracking_field"
+description: |-
+  Manages a work item field in Azure DevOps.
+---
+
+# azuredevops_workitemtracking_field
+
+Manages a work item field in Azure DevOps.
+
+~> **Note:** Custom fields are created at the organization level, not the project level. The Azure DevOps API does not support project-scoped custom fields.
```
magodo, line 12: Seems the project scoped fields are also supported by the API/SDK. Could you double confirm?

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/acceptancetests/testutils/field.go
Fresa: done

azuredevops/internal/acceptancetests/testutils/field.go
Fresa: I reran these tests a bunch of times without the retry and the specific error check, and I cannot reproduce the issue it solved, so I removed it.

azuredevops/internal/acceptancetests/resource_workitemtracking_field_test.go
Fresa: these could have the wrong values set without import failing due to the resource having an unexpected behavior implemented. These asserts guard against unexpected behavioral changes (the implementation changes) while being explicit about the expected behavior

website/docs/r/workitemtracking_field.html.markdown
Fresa: I agree, I removed them!

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: nice catch, done!

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: done

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: done

azuredevops/internal/acceptancetests/resource_workitemtracking_field_test.go
Fresa: After some thinking, I tend to agree with you. The only argument I have against removing these assertions is that they provide explicit expected behavior, which is a bit simpler to understand, than having to know that the test framework will assert that internally, except when the attribute is compute only.

Btw, the check step also diffs, so the import step isn't needed just to get that diff, but it's useful to assert that create/update doesn't diff against import.

I'll remove it as per your suggestion.

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: It's described here: https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/customize-process-field?view=azure-devops#delete-a-custom-field

website/docs/r/workitemtracking_field.html.markdown
Fresa: I actually posted a question on Developer Community regarding it: https://developercommunity.visualstudio.com/t/Creating-custom-fields-scoped-to-project/11021951#T-ND11022056

azuredevops/internal/service/workitemtracking/resource_field.go
magodo: From the document:

> Deleting a field removes all associated data, including historical values.

Seems undelete is mainly to recover these "associated data", which seeem not managed by terraform. In this case, it makes more sense to use other way to "undelete" the fields, instead of introducing this attribute in the provider?

website/docs/r/workitemtracking_field.html.markdown
magodo: Great, can you add a dummy property `project_id` and comment it out in the resource schema definition, then put this link above this property as a comment to describe why it is not supported?

azuredevops/internal/acceptancetests/resource_workitemtracking_field_test.go
magodo: One thing you'll need to check for each step after apply is the resource does exist, example: https://github.com/microsoft/terraform-provider-azuredevops/blob/9f998837fb362d6a9303a2f555244a1cd85b870e/azuredevops/internal/acceptancetests/resource_git_repository_test.go#L400-L423

Fresa: that's fair, I added an assertion that the id get's computed, it will test that the node exists as well. The import step also verifies this, assuming there is an import step :)

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: I interpret this as it will try to restore all fields in the work items of this type that also got deleted when deleting the WIT (the dataplane of the field instances). The mentioned history is related to the history of the work items I think (but for the field in this context), which is available in the history tab on each work item. 
image

In comparison, looking at [removing field from the wit](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/customize-process-field?view=azure-devops#remove-a-custom-field-from-a-form), it seems that operation doesn't destroy any data.

I haven't really tried this functionality so I'm not sure what else it might effect, controls referencing a field for example, are those also deleted/restored?

website/docs/r/workitemtracking_field.html.markdown
Fresa: done

Fresa:
@magodo ping, just a friendly reminder, when you have the time :)

azuredevops/internal/service/workitemtracking/resource_field.go
magodo: Since this is field only affect the "Create" behavior and it won't be returned by the API response, it look like a good fit to mark it as a [write-only attribute](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments).

magodo:
@Fresa One last comment is to mark the `restore` as a [write-only attribute](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments). Other than that, this PR LGTM! Thanks!

azuredevops/internal/service/workitemtracking/resource_field.go
Fresa: good idea, done!

Fresa:
@magodo I've added tests for every field type, turns out they behave differently, and I found what I think is a bug in the downstream API for identity fields. I've documented it in the resource.

magodo (approved):
LGTM!

```shell
terraform-provider-azuredevops on  add-fields via 🐹 v1.25.5
💤 TF_ACC=1 go test -v -run='TestAccWorkItemTrackingField_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItemTrackingField_Basic
=== PAUSE TestAccWorkItemTrackingField_Basic
=== RUN   TestAccWorkItemTrackingField_Complete
=== PAUSE TestAccWorkItemTrackingField_Complete
=== RUN   TestAccWorkItemTrackingField_Boolean
=== PAUSE TestAccWorkItemTrackingField_Boolean
=== RUN   TestAccWorkItemTrackingField_Html
=== PAUSE TestAccWorkItemTrackingField_Html
=== RUN   TestAccWorkItemTrackingField_Integer
=== PAUSE TestAccWorkItemTrackingField_Integer
=== RUN   TestAccWorkItemTrackingField_DateTime
=== PAUSE TestAccWorkItemTrackingField_DateTime
=== RUN   TestAccWorkItemTrackingField_PlainText
=== PAUSE TestAccWorkItemTrackingField_PlainText
=== RUN   TestAccWorkItemTrackingField_Double
=== PAUSE TestAccWorkItemTrackingField_Double
=== RUN   TestAccWorkItemTrackingField_Identity
=== PAUSE TestAccWorkItemTrackingField_Identity
=== RUN   TestAccWorkItemTrackingField_TreePath
=== PAUSE TestAccWorkItemTrackingField_TreePath
=== RUN   TestAccWorkItemTrackingField_History
=== PAUSE TestAccWorkItemTrackingField_History
=== RUN   TestAccWorkItemTrackingField_Guid
=== PAUSE TestAccWorkItemTrackingField_Guid
=== RUN   TestAccWorkItemTrackingField_Lock
=== PAUSE TestAccWorkItemTrackingField_Lock
=== RUN   TestAccWorkItemTrackingField_Restore
=== PAUSE TestAccWorkItemTrackingField_Restore
=== CONT  TestAccWorkItemTrackingField_Basic
=== CONT  TestAccWorkItemTrackingField_Integer
=== CONT  TestAccWorkItemTrackingField_Double
=== CONT  TestAccWorkItemTrackingField_Boolean
=== CONT  TestAccWorkItemTrackingField_Guid
=== CONT  TestAccWorkItemTrackingField_Lock
=== CONT  TestAccWorkItemTrackingField_Complete
=== CONT  TestAccWorkItemTrackingField_Html
=== CONT  TestAccWorkItemTrackingField_Restore
=== CONT  TestAccWorkItemTrackingField_Identity
=== CONT  TestAccWorkItemTrackingField_TreePath
=== CONT  TestAccWorkItemTrackingField_History
--- PASS: TestAccWorkItemTrackingField_TreePath (5.12s)
=== CONT  TestAccWorkItemTrackingField_DateTime
--- PASS: TestAccWorkItemTrackingField_Identity (5.12s)
=== CONT  TestAccWorkItemTrackingField_PlainText
--- PASS: TestAccWorkItemTrackingField_Basic (5.12s)
--- PASS: TestAccWorkItemTrackingField_History (5.14s)
--- PASS: TestAccWorkItemTrackingField_Html (5.16s)
--- PASS: TestAccWorkItemTrackingField_Complete (5.17s)
--- PASS: TestAccWorkItemTrackingField_Guid (5.23s)
--- PASS: TestAccWorkItemTrackingField_Boolean (5.31s)
--- PASS: TestAccWorkItemTrackingField_Double (5.39s)
--- PASS: TestAccWorkItemTrackingField_Integer (5.42s)
--- PASS: TestAccWorkItemTrackingField_Lock (6.66s)
--- PASS: TestAccWorkItemTrackingField_DateTime (1.55s)
--- PASS: TestAccWorkItemTrackingField_PlainText (1.76s)
--- PASS: TestAccWorkItemTrackingField_Restore (8.06s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	8.066s
```

PR #1477, opened 2026-01-02
New resource: `azuredevops_workitemtrackingprocess_field`

Fresa:
## All Submissions:

Adds a new field resource that mangages [fields](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/fields?view=azure-devops-rest-7.1) for a work item type. 

* `allowedValues` is not included, as it doesn't seem to work. After some trial and error in the UI, I think this might have been superseeded by the [lists](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/lists?view=azure-devops-rest-7.1) api.
* `defaultValue` is typed as an object, but the downstream API returns all kinds of weird errors if one tries to send in anything else than a string. Therefor it is defined as a string in the resource.

**NOTE:** This PR is based on [1475](https://github.com/microsoft/terraform-provider-azuredevops/pull/1475), which should be reviewed first.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


Fresa:
@magodo this one is ready for review, when you have the time!

azuredevops/internal/service/workitemtracking/retry.go
```diff
@@ -0,0 +1,12 @@
+package workitemtracking
+
+import (
+	"context"
+	"time"
+
+	"github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/service/workitemtrackingprocess"
+)
+
+func RetryOnUnexpectedException(ctx context.Context, timeout time.Duration, f func() error) error {
```
Fresa: I made a simple adapter here between these two packages, but maybe all these retry methods should be moved to the utils package instead?

azuredevops/internal/service/workitemtracking/resource_field.go
```diff
@@ -373,7 +373,9 @@ func resourceFieldDelete(ctx context.Context, d *schema.ResourceData, m interfac
 		FieldNameOrRefName: &referenceName,
 	}
 
-	err := clients.WorkItemTrackingClient.DeleteWorkItemField(clients.Ctx, args)
+	err := RetryOnUnexpectedException(ctx, d.Timeout(schema.TimeoutDelete), func() error {
```
Fresa: I added this here in this PR because I never saw this until I added the acceptances tests for the workitemtrackingprocess_field resource

website/docs/r/workitemtrackingprocess_field.html.markdown
```diff
@@ -0,0 +1,98 @@
...
+resource "azuredevops_workitemtracking_field" "example" {
+  name           = "Priority Level"
+  reference_name = "Custom.PriorityLevel"
+  type           = "string"
+}
+
+resource "azuredevops_workitemtrackingprocess_field" "example" {
+  process_id              = azuredevops_workitemtrackingprocess_process.example.id
+  work_item_type_ref_name = azuredevops_workitemtrackingprocess_workitemtype.example.reference_name
+  reference_name          = azuredevops_workitemtracking_field.example.reference_name
```
magodo: Maybe we shall simply make them `work_item_type_id` and `field_id` for a better UX?

website/docs/r/workitemtrackingprocess_field.html.markdown
```diff
@@ -0,0 +1,98 @@
...
+
+In addition to the Arguments listed above - the following Attributes are exported:
+
+* `id` - The reference name of the field.
+
+* `name` - The name of the field.
+
+* `type` - The type of the field.
+
+* `description` - The description of the field.
```
magodo: These seem useless as they are already exposed by the field resource/data source itself.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
```diff
@@ -0,0 +1,278 @@
...
+				Computed:    true,
+				Description: "The type of the field.",
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description of the field.",
+			},
+			"default_value": {
+				Type:        schema.TypeString,
```
magodo, line 56: This is string per your test. Was that because the field's type is defined as string hence you can only specify a default string value. Otherwise, if the field's type is other type, you shall specify a different default value type?

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
```diff
@@ -0,0 +1,278 @@
...
+				Default:     false,
+				Description: "If true, the field cannot be edited.",
+			},
+			"required": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "If true, the field cannot be empty.",
+			},
+			// We set this to write-only to circumvent this bug: https://developercommunity.visualstudio.com/t/Custom-field-APIs-are-missing-attributes/11032086
```
magodo, line 72: The downside of doing so is that the user won't be able to update this attribute alone due to the nature of write only attribute.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
```diff
@@ -0,0 +1,278 @@
...
+
+	processId := d.Get("process_id").(string)
+	witRefName := d.Get("work_item_type_ref_name").(string)
+	referenceName := d.Get("reference_name").(string)
+
+	fieldRequest := &workitemtrackingprocess.AddProcessWorkItemTypeFieldRequest{
+		ReferenceName: &referenceName,
+		ReadOnly:      converter.Bool(d.Get("read_only").(bool)),
+		Required:      converter.Bool(d.Get("required").(bool)),
+		AllowGroups:   converter.Bool(d.Get("allow_groups").(bool)),
```
magodo: Should we conditionally set this as it is optional (wo a default value)?

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
```diff
@@ -0,0 +1,278 @@
...
+	createdField, err := clients.WorkItemTrackingProcessClient.AddFieldToWorkItemType(ctx, args)
+	if err != nil {
+		return diag.Errorf("adding field %s to work item type %s: %+v", referenceName, witRefName, err)
+	}
+
+	if createdField.ReferenceName == nil {
+		return diag.Errorf("created field has no reference name")
+	}
+
+	d.SetId(*createdField.ReferenceName)
```
magodo: Arguably, we shall set the id as the format for the import since this resource is an association resource.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
```diff
@@ -0,0 +1,278 @@
...
+
+	processId := d.Get("process_id").(string)
+	witRefName := d.Get("work_item_type_ref_name").(string)
+	fieldRefName := d.Id()
+
+	fieldUpdate := &workitemtrackingprocess.UpdateProcessWorkItemTypeFieldRequest{
+		ReadOnly:    converter.Bool(d.Get("read_only").(bool)),
+		Required:    converter.Bool(d.Get("required").(bool)),
+		AllowGroups: converter.Bool(d.Get("allow_groups").(bool)),
+	}
```
magodo, line 202: Can we conditionally set these based on `d.HasChange()`, to allow use of `ignore_changes`?

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_field_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testutils.PreCheck(t, nil) },
+		ProviderFactories: testutils.GetProviderFactories(),
+		CheckDestroy:      checkProcessAndFieldDestroyed,
+		Steps: []resource.TestStep{
+			{
+				Config: basicField(workItemTypeName, processName, fieldName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(tfNode, "id"),
```
magodo, line 26: Shall we also check the other computed only attributes?

azuredevops/internal/utils/HttpResponse.go
magodo, line 11: Can we send the actual expected status code from the function call site?

azuredevops/internal/utils/tfhelper/tfhelper.go
```diff
@@ -72,6 +72,21 @@ func ParseImportedName(id string, expectedFormat string) (string, string, error)
 	return part1, part2, nil
 }
 
+// ParseImportedNameParts parses an imported ID into the expected number of parts separated by /.
+// Returns an error if the ID doesn't have exactly the expected number of non-empty parts.
+func ParseImportedNameParts(id string, expectedFormat string, expectedParts int) ([]string, error) {
```
magodo, line 77: This is unncessary, you can simply just use `strings.Split()` at the call site instead.

azuredevops/internal/service/workitemtracking/retry.go
magodo: Yes please. Otherwise, there will be risk to cause package cyclic dependencies.

magodo (changes requested):
Thanks for this PR!

website/docs/r/workitemtrackingprocess_field.html.markdown
Fresa: it's a good idea, it also hints that the "id" of the parent resource should be used as reference in order to couple their lifecycles. 

Done.

website/docs/r/workitemtrackingprocess_field.html.markdown
Fresa: Yeah I tend to agree, I've followed the downstream API here assuming there is a reason they are included. There is no data source for `azuredevops_workitemtracking_field` today though, but that could be added should there be a need.

Shall I remove them?

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: That was how I [interpret the API](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/fields/add?view=azure-devops-rest-7.1&tabs=HTTP#request-body) first as well considering the field is defined as type `object`, but it turns out the API expects the default value to always be a string no matter the type of the field. I used a json marshall approach first: https://github.com/Fresa/terraform-provider-azuredevops/commit/138f30b1391f3dbc76fdc4c76d096a3a1ba487ff, check the commit message for more info

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: yeah, any suggestions? A user could work around it by also changing something non-functional, like the description for example

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: makes sense semantically, done.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: that's fair, done.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: does that matter? I was of the impression that the SDK handles that (ignores configuration and reads the state instead). If it does matter, this problem exists on every resource I've implemented...

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_field_test.go
Fresa: we don't know whether other computed attributes will be set or not

azuredevops/internal/utils/HttpResponse.go
Fresa: This method contains the expected "status codes". It's used to determine if an an error response states that a resource doesn't exist. It's used everywhere when creating, deleting and reading

azuredevops/internal/utils/tfhelper/tfhelper.go
Fresa: It's a complement to the ParseImportedName helper, which only works for two-parts ID's. This one works with any parts. But yeah, the same functionality can be copy/pasted into every resource that imports multipart IDs. Should each resource handle this by them selves?

azuredevops/internal/service/workitemtracking/retry.go
Fresa: Done

website/docs/r/workitemtrackingprocess_field.html.markdown
magodo: Yes, please

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
magodo: If the type is string, then you don't need to marshal it as a JSON string, just a string literal shall work. Otherwise, if it is another type (e.g. object), then you'll need to marshal it as a JSON object?

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
magodo: Let's keep it as is and add a callout in the document in case a user want to update it.

azuredevops/internal/utils/tfhelper/tfhelper.go
magodo: Fair, let's keep this.

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: I only got the API to accept default values formatted as literal strings. I first tried to represent default values as the actual type, i.e. 42 if the field was of type integer instead of "42", using the `any` go type, that didn't work, the api responded with errors that it couldn't parse it. Unfortunately I never committed that, so there is no history of me trying. 

Then I went with the stringifed JSON approach, i.e. "42" for `integer`, "\"42\"" for `string`, That also caused the API to respond with errors, so I changed default values to always send a string literal, i.e. "42" for `integer`, and "42" for `string`, and that seems to be the only thing the API accepts. 

I just added another test for integer typed fields just to visualize that this works. Seems the API parses and validates the string literal sent according to the actual field type. If a non-integer like default value is sent for an integer typed field, it will respond with something like:
`Error: adding field Custom.testaccf1141vzbb2 to work item type test-acc-rt1bq49afu.testaccz0e8mc6d3h: VS402809: foo is an invalid value for field type Integer. Change the field value and try again.`

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
Fresa: Done

website/docs/r/workitemtrackingprocess_field.html.markdown
Fresa: Done

azuredevops/internal/service/workitemtrackingprocess/resource_field.go
magodo: Thanks for the additional context!

magodo (approved):
LGTM! Thank you!

## Test

```shell
💤 TF_ACC=1 go test -v -run='TestAccWorkitemtrackingprocessField_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkitemtrackingprocessField_Basic
=== PAUSE TestAccWorkitemtrackingprocessField_Basic
=== RUN   TestAccWorkitemtrackingprocessField_Identity
=== PAUSE TestAccWorkitemtrackingprocessField_Identity
=== RUN   TestAccWorkitemtrackingprocessField_Integer
=== PAUSE TestAccWorkitemtrackingprocessField_Integer
=== RUN   TestAccWorkitemtrackingprocessField_Update
=== PAUSE TestAccWorkitemtrackingprocessField_Update
=== CONT  TestAccWorkitemtrackingprocessField_Basic
=== CONT  TestAccWorkitemtrackingprocessField_Integer
=== CONT  TestAccWorkitemtrackingprocessField_Update
=== CONT  TestAccWorkitemtrackingprocessField_Identity
--- PASS: TestAccWorkitemtrackingprocessField_Identity (4.68s)
--- PASS: TestAccWorkitemtrackingprocessField_Integer (4.78s)
--- PASS: TestAccWorkitemtrackingprocessField_Basic (5.80s)
--- PASS: TestAccWorkitemtrackingprocessField_Update (6.31s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	6.317s
```

PR #1482, opened 2026-01-08
Add pages to work item type layout

Fresa:
## All Submissions:

Add new resource for work item type [pages](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/pages?view=azure-devops-rest-7.1)

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


Fresa:
@magodo This one is ready for review, when you have the time!

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
```diff
@@ -0,0 +1,247 @@
...
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Computed:    true,
+				Description: "Order in which the page should appear in the layout.",
+			},
+			"visible": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "A value indicating if the page should be hidden or not.",
```
magodo:
```suggestion
				Description: "A value indicating if the page should be visible or not.",
```

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+			"section": {
```
magodo: Can we make this a plural as it is computed only:

```suggestion
			"sections": {
```

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
```diff
@@ -0,0 +1,247 @@
...
+	return []*schema.ResourceData{d}, nil
+}
+
+func createResourcePage(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	clients := m.(*client.AggregatedClient)
+
+	page := workitemtrackingprocess.Page{
+		Label:    converter.String(d.Get("label").(string)),
+		Visible:  converter.Bool(d.Get("visible").(bool)),
+		PageType: &workitemtrackingprocess.PageTypeValues.Custom,
```
magodo, line 103: I'm not familiar with this resource, whilst by fixing this does it mean the other variants of the page are not a terraform resource candidate?

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	rawConfig := d.GetRawConfig().AsValueMap()
```
magodo: We shall use d.GetRawPlan() instead.

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
```diff
@@ -0,0 +1,247 @@
...
+	pageId := d.Id()
+	processId := d.Get("process_id").(string)
+	witRefName := d.Get("work_item_type_id").(string)
+
+	updatePage := &workitemtrackingprocess.Page{
+		Id:      &pageId,
+		Label:   converter.String(d.Get("label").(string)),
+		Visible: converter.Bool(d.Get("visible").(bool)),
+	}
+	rawConfig := d.GetRawConfig().AsValueMap()
```
magodo: Same here

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
Fresa: done

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
Fresa: A custom page type is the only fully manageable page type (create/delete/update). The other three, History, Links and Attachments are system controlled pages; the system provisions them for all the work item types, and AFAIK they cannot be customized nor can they be created. I'm not really sure PageType here even does anything other than show intent, the API endpoint here will always return Custom no matter what, at least today.

Then there are inherited pages, the "Details" page for example on all inherited work item types. They are provisioned along side the work item type and only their label can be modified. I've [created a separate resource](https://github.com/microsoft/terraform-provider-azuredevops/pull/1492) to manage them, as their modification process differ quite a lot from custom pages, alas they do use the same API endpoints.

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
Fresa: I wrote a long explanation why I chose GetRawConfig + Get, until I eventually realized that plan is much simpler and deals with all edge cases that I can think off. Nice catch! We should probably do this for all other resources that also declares order in a similar way.

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
Fresa: done!

azuredevops/internal/service/workitemtrackingprocess/resource_page.go
magodo: Thanks for the additional context! Then do you think it make more sense to rename this resource to xxx_custom_page?

Fresa: We would need to do the same for controls to be consistent, which might be difficult since we've merged it. The  documentation does call it "[custom page](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/customize-process-form?view=azure-devops#add-a-custom-page)" though, so it's not a bad idea per-se. 

Work Item Types are btw also mentioned as ["custom" and "inherited"](https://learn.microsoft.com/en-us/azure/devops/organizations/settings/work/inheritance-process-model?view=azure-devops&tabs=agile-process#work-item-type-customizations) along with many other resources; controls, groups, fields, processes, but it's also mixed with the word "inherited".

One could argue that "inherited page" isn't a good name either as it's defined as a "custom" page by the API with a couple of other attributes set. There are even "custom" "inherited" pages tagging along when creating a work item type that doesn't inherit from a system defined WIT but can be "overridden", so `custom_inherited_overridden_page`? :)

magodo: Fair enough, then let's keep the name as is.

magodo (approved):
LGTM! Thanks!

```shell
terraform-provider-azuredevops on  add-pages via 🐹 v1.25.4
💤 TF_ACC=1 go test -v -run='TestAccWorkitemtrackingprocessPage_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkitemtrackingprocessPage_Basic
=== PAUSE TestAccWorkitemtrackingprocessPage_Basic
=== RUN   TestAccWorkitemtrackingprocessPage_Update
=== PAUSE TestAccWorkitemtrackingprocessPage_Update
=== CONT  TestAccWorkitemtrackingprocessPage_Basic
=== CONT  TestAccWorkitemtrackingprocessPage_Update
--- PASS: TestAccWorkitemtrackingprocessPage_Basic (14.14s)
--- PASS: TestAccWorkitemtrackingprocessPage_Update (16.31s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        16.333s
```

Issue #1488, opened 2026-01-13
Manage inherited layout resources for work item type


Fresa:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
When creating an inherited work item type, the layout is inherited from the parent work item type. The layout consists of pages, sections, groups and controllers. These have limited customizations, controls can for example only change visibility and the label. 

These resources are not manageable currently. When implementing this, I had the vision that the user could utilize Terraform's `import` directive to manage these few customizations, but I realize that terraform doesn't [support values in the id directive that are unknown during plan](https://github.com/hashicorp/terraform/issues/34152). 

I'm thinking that we could perhaps manage this in the wit resource instead.

There is already the computed [`pages` attribute](https://github.com/microsoft/terraform-provider-azuredevops/blob/0b20424c10c56d66d9def7b0112653ec829b4e77/azuredevops/internal/service/workitemtrackingprocess/resource_work_item_type.go#L98), that returns a subset of data for all layout elements, inherited or not. So that won't work. Should we perhaps add another list/map attribute, `inherited_pages`, which would be optional, where it's possible to manage the inherited resources and their specific mutable attributes?

## Mutable Inherited Attributes
* pages: label
* controls: label, visible

I think these are the only ones.

Thoughts?

### New or Affected Resource(s)
* azuredevops_workitemtrackingprocess_workitemtype

### Potential Terraform Configuration

```hcl
resource "azuredevops_workitemtrackingprocess_workitemtype" "user_story" {
  process_id  = azuredevops_workitemtrackingprocess_process.foo.id
  name        = "User Story"

  parent_work_item_reference_name = "Microsoft.VSTS.WorkItemTypes.UserStory"

  inherited_pages = {
    "Agile.User Story.User Story" = {
      visible = true # page visibility

      inherited_groups = {
        "Agile.User Story.User Story.Acceptance Criteria.WideGroup" = {
          inherited_controls = {
            "Microsoft.VSTS.Common.AcceptanceCriteria" = {
              visible = false # hide this control
              label   = "Foo"  # change the label
            }
          }
        }
      }
    }

[...]

PR #1493, opened 2026-01-19
New resource: `azuredevops_workitemtrackingprocess_inherited_state`

Fresa:
## All Submissions:

Add a new resource `azuredevops_workitemtrackingprocess_inherited_state`, that manages inherited work item type [state](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/states?view=azure-devops-rest-7.1) modifications.
image

image

Issue Number: #1488 

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

N/A

## Other information


azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
```diff
@@ -0,0 +1,233 @@
...
+				Description:      "The ID of the process.",
+			},
+			"work_item_type_id": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
+				Description:      "The ID (reference name) of the work item type.",
+			},
+			"name": {
```
Fresa, line 47: We could use the ID here instead if we like. I chose name because it's simpler to identify than a UUID. State names are always unique within a WIT.

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
```diff
@@ -0,0 +1,233 @@
...
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotWhiteSpace),
+				Description:      "Name of the inherited state to manage.",
+			},
+			"visible": {
+				Type:        schema.TypeBool,
+				Optional:    true,
```
Fresa, line 56: This could have been configured as required, but I chose to align it with all the other resources that have visibility configurable.

Fresa:
@magodo this PR is ready for review, when you have the time!

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil, err
+	}
+
+	if states != nil {
+		for _, state := range *states {
+			if state.Name != nil && *state.Name == name {
+				if state.CustomizationType == nil {
+					return nil, fmt.Errorf("state %q has no customization type", name)
+				}
+				if *state.CustomizationType == workitemtrackingprocess.CustomizationTypeValues.Custom {
```
magodo, line 226: Shouldn't we look for a state whose CustomizationType equals to Inherited?

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func updateResourceInheritedState(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	clients := m.(*client.AggregatedClient)
+
+	stateId := d.Id()
+	processId := d.Get("process_id").(string)
+	witRefName := d.Get("work_item_type_id").(string)
+
+	visible, err := getBoolAttributeFromConfig(d, "visible")
```
magodo, line 157: Maybe we shall consider to modify this function `getBoolAttributeFromConfig` to a generic function that works for other types than boolean? Though it shall be done in another PR.

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
magodo: Since the other related resources all use id, shall we also use id here to be consistent?

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
```diff
@@ -0,0 +1,233 @@
...
+				WitRefName:     &witRefName,
+				StateId:        converter.UUID(stateId),
+				HideStateModel: &workitemtrackingprocess.HideStateModel{Hidden: &hidden},
+			})
+			if err != nil {
+				return diag.Errorf("hiding state: %+v", err)
+			}
+		} else {
+			// Use DELETE to unhide a state
+			err := clients.WorkItemTrackingProcessClient.DeleteStateDefinition(ctx, workitemtrackingprocess.DeleteStateDefinitionArgs{
```
magodo, line 176: Assuming deleting it here won't really delete the state definition but just revert it, and it won't impact a follow up `GetStateDefinition` call?

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_inherited_state_test.go
```diff
@@ -0,0 +1,209 @@
...
+			StateId:    &stateUUID,
+		})
+		if err != nil {
+			return fmt.Errorf("getting state definition: %w", err)
+		}
+
+		if state == nil {
+			return fmt.Errorf("inherited state should still exist after removing from Terraform but it was not found")
+		}
+
```
magodo, line 193: Shall we also test the visibility here is null once it was deleted?

magodo (changes requested):
Thanks for this PR!

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
Fresa: "Inherited" states are actually labeled as "System" for some reason, when hiding a "System" state it turns into an "Inherited" state. The [Hide operation](https://learn.microsoft.com/en-us/rest/api/azure/devops/processes/states/hide-state-definition?view=azure-devops-rest-7.1&tabs=HTTP) only works on "System" states. I've never seen an "Inherited" state on an inherited work item type, except when hiding it. One could argue that we should call it `azuredevops_workitemtrackingprocess_system_state` instead, I'm not sure what makes most sense since it switches between them. System/inherited states only seem to exist on inherited work item types, and the description in the States tab in the UI (see screenshot in PR description above) refers to it as "inherited" state, probably because it is in fact inherited from the parent WIT.

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
Fresa: It would simplify the implementation a bit as well. The IDs seems to be hardcoded globally, so predictable-ish, but would still require the user to find it via the state list endpoint or WIT GET endpoint. I don't have a strong opinion for either, you choose!

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
Fresa: Yepp, exactly. 

If trying to use the Hide endpoint with `Hidden: false`, it will error and explicitly suggest using the DELETE endpoint for reverting.

azuredevops/internal/acceptancetests/resource_workitemtrackingprocess_inherited_state_test.go
Fresa: Good point, done!

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
magodo: Well, thanks for the context! Then could you add some comments here to clarify?

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
magodo: ~With a further look, I would suggest we expand the `azuredevops_work_item_type` to expose the `states` (similar as the `pages`). So that the user can further convert the array of `states` to a map keyed by the state name, then reference the state id by the name. WDYT?~

For the UX, maybe we shall just keep the `name`, which the user can easily learn from the document/UI?

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
Fresa: done!

azuredevops/internal/service/workitemtrackingprocess/resource_inherited_state.go
Fresa: I think that makes sense

magodo (approved):
LGTM!

```shell
terraform-provider-azuredevops on  manage-inherited-states via 🐹 v1.25.6
💤 TF_ACC=1 go test -v -run='TestAccWorkitemtrackingprocessInheritedState_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkitemtrackingprocessInheritedState_Basic
=== PAUSE TestAccWorkitemtrackingprocessInheritedState_Basic
=== RUN   TestAccWorkitemtrackingprocessInheritedState_Update
=== PAUSE TestAccWorkitemtrackingprocessInheritedState_Update
=== RUN   TestAccWorkitemtrackingprocessInheritedState_RemoveFromState
=== PAUSE TestAccWorkitemtrackingprocessInheritedState_RemoveFromState
=== CONT  TestAccWorkitemtrackingprocessInheritedState_Basic
=== CONT  TestAccWorkitemtrackingprocessInheritedState_RemoveFromState
=== CONT  TestAccWorkitemtrackingprocessInheritedState_Update
--- PASS: TestAccWorkitemtrackingprocessInheritedState_Basic (3.72s)
--- PASS: TestAccWorkitemtrackingprocessInheritedState_Update (3.94s)
--- PASS: TestAccWorkitemtrackingprocessInheritedState_RemoveFromState (5.62s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        5.624s
```

PR #1505, opened 2026-02-09
Adding an optional, default false parameter to skip input configuration validation for service endpoints using the generic_v2 resource and adding types data source.

aklakina:
## All Submissions:

The `resource_serviceendpoint_generic_v2` had pre-validation of incoming configuration which tried to interpret the input descriptors of service endpoint type. Sadly this functionality does not work perfectly so added a switch to enable it if needed but default stays disabled. Also implemented 2 data sources which are supposed to query available type of service endpoints and their definitions.

Issue Number: -

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


aklakina:
Hi @magodo 

I think the pipeline failures are false negatives here. Can you restart them please? Also this PR is ready for review.

magodo:
Hi @aklakina, sorry for the delay on reviewing this PR. To make the PR self contained, can we consider to split it into different PRs for different purposes? For the toggle of validation on the generic v2 resource, what is the case that it wants to be disabled?

aklakina:
Hi @magodo,

Sure, I'll split this into a feature and fix prs.

As for why it is disabled, the parameters inside the service connections and the authentication parameters as well, all have a field called `condition`. My implementation did not take that field into account which meant the validation of the input data failed yet the input data did have everything necessary to create the service connection. Here I'm specifically talking about the azurerm connection, but there can be other complex service connections too.

And since the validation itself can only run normally in apply-time because of undeterministic input values, it is almost the same as letting the api validate the object. I could also remove the code that was responsible for this validation.

magodo:
If that is apply-time only, then I agree we shall remove the validation and just let the API fail. On the other hand, if those values are statically known at plan-time, validating them at the `CustomizeDiff` might still benefit the users to avoid break in the middle of an apply.

PR #1516, opened 2026-02-27
Add attributes to ClassificationNodeSchema

brittandeyoung:
## All Submissions:

## Description

Adds the `attributes` field to the `azuredevops_area` and `azuredevops_iteration` data sources. This will allow reading the start and end dates for an iteration.

- [ ] Yes
- [x] No

## Test Result

I did not see any integration tests related to this datasource as there is not a corresponding resource yet. 

## Related Issue(s)

Fix #0000


azuredevops/internal/service/workitemtracking/utils/classification.go
```diff
@@ -67,6 +67,10 @@ func CreateClassificationNodeSchema(outer map[string]*schema.Schema) map[string]
 				},
 			},
 		},
+		"attributes": {
+			Type:     schema.TypeMap,
```
magodo, line 71: Can we also specify the `Elem` here?

azuredevops/internal/service/workitemtracking/utils/classification.go
```diff
@@ -116,6 +120,10 @@ func ReadClassificationNode(clients *client.AggregatedClient, d *schema.Resource
 		d.Set("has_children", converter.ToBool(node.HasChildren, false))
 	}
 
+	if node.Attributes != nil {
+		d.Set("attributes", *node.Attributes)
```
magodo, line 124: The type of `*node.Attributes` is `map[string]any`, which wants to be casted to the actual type. Also we shall handle the potential error returned by the `d.Set()` here.

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline. Besides, can you update and pass the acceptance tests?

PR #1521, opened 2026-03-08
azuredevops_workitem: Add description and additional_fields_json parameters 

### LINKED ISSUE #1520: `azuredevops_workitem` resource missing `story_points`, `description` and `acceptance_criteria` fields
### Community Note

### Description

Currently the `custom_fields` field on the `azuredevops_workitem` prepends the `Custom` prefix. This prevents this from being used for additional fields not part of this resource like the following:

* `Microsoft.VSTS.Scheduling.StoryPoints`
* `Microsoft.VSTS.Common.AcceptanceCriteria`
* `System.Description`

Our implementation requires these fields which makes us unable to use this terraform resource. 

* azuredevops_workitem

```hcl
resource "azuredevops_workitem" "example" {
  project_id = data.azuredevops_project.example.id
  title      = "Example Work Item"
  type       = "Issue"
  state      = "Active"
  tags       = ["Tag"]
  story_points = 5.0
  description = "This is a description"
  acceptance_criteria = "This feature must do this thing"
}
```

### References

* #0000

brittandeyoung:
## All Submissions:

## Description

Adds the following optional fields to the `azuredevops_workitem` resource:

1. `description`
2. `story_points` 
3. `acceptance_criteria`

- [ ] Yes
- [x] No

## Test Result
```
➜ terraform-provider-azuredevops (feature/workitem-custom-fields-remove-prefix) ✔ TF_ACC=1 go test -v -run='TestAccWorkItem_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItem_basic
=== PAUSE TestAccWorkItem_basic
=== RUN   TestAccWorkItem_titleUpdate
=== PAUSE TestAccWorkItem_titleUpdate
=== RUN   TestAccWorkItem_tagUpdate
=== PAUSE TestAccWorkItem_tagUpdate
=== RUN   TestAccWorkItem_parent
=== PAUSE TestAccWorkItem_parent
=== RUN   TestAccWorkItem_parentUpdate
=== PAUSE TestAccWorkItem_parentUpdate
=== RUN   TestAccWorkItem_parentDelete
=== PAUSE TestAccWorkItem_parentDelete
=== RUN   TestAccWorkItem_additionalFieldsJson
=== PAUSE TestAccWorkItem_additionalFieldsJson
=== RUN   TestAccWorkItem_description
=== PAUSE TestAccWorkItem_description
=== CONT  TestAccWorkItem_basic
=== CONT  TestAccWorkItem_parentUpdate
=== CONT  TestAccWorkItem_tagUpdate
=== CONT  TestAccWorkItem_parent
=== CONT  TestAccWorkItem_titleUpdate
=== CONT  TestAccWorkItem_additionalFieldsJson
=== CONT  TestAccWorkItem_description
=== CONT  TestAccWorkItem_parentDelete
--- PASS: TestAccWorkItem_basic (39.05s)
--- PASS: TestAccWorkItem_parent (40.91s)
--- PASS: TestAccWorkItem_additionalFieldsJson (40.96s)
--- PASS: TestAccWorkItem_titleUpdate (42.07s)
--- PASS: TestAccWorkItem_parentUpdate (43.26s)
--- PASS: TestAccWorkItem_parentDelete (43.46s)
--- PASS: TestAccWorkItem_tagUpdate (46.79s)
--- PASS: TestAccWorkItem_description (48.49s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	49.678s
```

## Related Issue(s)

Fix #1520


magodo:
@brittandeyoung Thanks for submitting this PR! My main concern for this method is it can make inflate the resource properties pretty soon. Instead, can we introduce a new map called `additional_fields` that deprecates/overrides the `custom_fields`. The key of the `additional_fields` will then hold the complete field path starting from `/fields/`, which include `Custom.`, `Microsoft.VSTS`, etc. The challenge is to introduce it in a non-breaking way, which means you'll need to keep both maps continue working, but only one of them can be explicitly specified by the user, while both will be populated (i.e. both have `Computed: true`).

brittandeyoung:
@magodo I had originally started down the path you are suggesting. The issue I ran into comes with importing the resource. As these additional fields key values will not be available at import time, it will cause a plan difference on import. After running a plan after import, the keys would be available to terraform and there would no longer be a difference. This could cause unexpected messaging to the end user.

This would really only be an issue when running the terraform import command instead of using the import block.

Is this a behavior the provider team is okay with?

If so I will be happy to update the pr to use the `additional_fields` method.

magodo:
@brittandeyoung The import stage should be able to read those fields back as part of the `Read()`, specifically in `flattenFields` these fields are set to the state. This shall end up without a plan diff.

brittandeyoung:
@magodo 

You can see that in the `flattenFields` function that we are filtering which `fields` are read from the API:

```go

var systemFieldMapping = map[string]string{
	"System.State":         "state",
	"System.Title":         "title",
	"System.WorkItemType":  "type",
	"System.AreaPath":      "area_path",
	"System.IterationPath": "iteration_path",
	"System.Parent":        "parent_id",
	"System.Description":   "description",
}

var fieldMapping = map[string]string{
	"title":               "System.Title",
	"type":                "System.WorkItemType",
	"state":               "System.State",
	"area_path":           "System.AreaPath",
	"iteration_path":      "System.IterationPath",
	"parent_id":           "System.Parent",
	"description":         "System.Description",
	"story_points":        "Microsoft.VSTS.Scheduling.StoryPoints",
	"acceptance_criteria": "Microsoft.VSTS.Common.AcceptanceCriteria",
}

var managedFieldMapping = map[string]string{
	"Microsoft.VSTS.Scheduling.StoryPoints":    "story_points",
	"Microsoft.VSTS.Common.AcceptanceCriteria": "acceptance_criteria",
}


func flattenFields(d *schema.ResourceData, m *map[string]interface{}) {
	customFields := make(map[string]interface{})
	for key, value := range *m {
		if v, ok := systemFieldMapping[key]; ok {
			d.Set(v, value)
		} else if v, ok := managedFieldMapping[key]; ok {
			d.Set(v, value)
		} else if strings.HasPrefix(key, customFieldsPrefix) {
			customFields[strings.ReplaceAll(key, customFieldsPrefix, "")] = value
		} else if "System.Tags" == key {
			d.Set("tags", strings.Split(value.(string), "; "))
		}
	}

	d.Set("custom_fields", customFields)
}
```

This filter is made of static field mappings. Without these static fields in the provider code, we would need to read ALL fields from the API. This would result in a permanent diff of the field unless ALL fields where specified in `additional_fields`

So we either need to have a static list inside the provider code of fields supported by the `additional_fields` configuration, or individuals would need to provide ALL fields into this map when in use which would override automatically set fields and cause issues like `Microsoft.VSTS.Common.StateChangeDate`

magodo:
Could you educate me what are these two `Microsoft.VSTS.` fields? Are they always available?

Besides, the way you implemented these attributes seems to be a breaking change as well as it will read these attributes (e.g. `description` assuming it is always available) even if the user didn't specify them in the config from the old version?

brittandeyoung:
@magodo  

The "Acceptance Criteria" field and "Description" field are documented here: https://learn.microsoft.com/en-us/azure/devops/boards/queries/titles-ids-descriptions?view=azure-devops

The "Story Points" Field is documented here: https://learn.microsoft.com/en-us/azure/devops/boards/queries/query-numeric?view=azure-devops#fields

According the the docs "Acceptance Criteria" and "Story Points" are not available on all work item types and is subject to project level configurations; however, from the tests that I wrote, you can set these fields in the API (and read them) even on unsupported Work Item types without error. This is why I added test cases with the "Issue" type knowing that these fields are not supported for that type. From what I can tell, these fields still set in the API, but are not used for the type (do not show up in the UI).

I ran all my tests against a brand new project.

I do not believe this is breaking as the code is configured to read the field and set it IF it exists. The value will simply be null if the field  is not returned by the API. 

I have extensive experience writing terraform providers and am willing to implement this in the preferred method of the provider team. I have seen some other provider teams move to a separate resource per item "type" to get around some of these "field is only valid for certain types" issue. For example, adding a `azuredevops_workitem_userstory` resource and a unique resource per type to allow adding fields specific to that type. 

please advise on how you would like me to proceed.

magodo:
@brittandeyoung Thank you for providing more context! For the `desription`, I have no problem about the current design as it is available for all work item types. For "Acceptance Criteria" and "Story Points", I noticed that they are similar to Custom fileds, except there is one `Microsoft.VSTS.Common.StateChangeDate` field which is always there. So maybe we shall still consider to introduce an `additional_fields` and explicitly ignore the `Microsoft.VSTS.Common.StateChangeDate` during flattening. Then it shall behave the same as `custom_fields`, while we can still separate the matter cleanly.

brittandeyoung:
@magodo I will get to work on updating this PR to fit that. I am concerned with their being different types required for fields. My thought to allow for that is that the `additional_fields_json` field would be json in string format (The AWS provider does this for some resources), then we use `jsonencode()` function. This allows for the user to pass in the type of the field with the json. Ill ping you once I have some working code

brittandeyoung:
@magodo I have updated the PR with the following (Keeping the `description`  argument):

1. Added the `additional_fields_json` argument that excepts JSON in string format. This will allow for adding any additional field using its required type.
2. Added a deprecation notice on the `custom_fields` argument informing users to use the `additional_fields_json` argument.
3. Added validation to ensure both `custom_fields` and `additional_fields_json` are not used together.
4. Updated the documentation for the `custom_fields` and `additional_fields_json` properties. 
5. Added an example for the `additional_fields_json`  argument to the docs.
6. Added acceptance tests for the `additional_fields_json` argument. 
7. Added WhitespaceJsonDifference function to the suppress package to help suppress plan difference due to whitespace in the JSON string field.
8 . Added Unit tests to validate the WhitespaceJsonDifference function.

```
After these changes ran the acceptance tests for the resource:
➜ terraform-provider-azuredevops (feature/workitem-custom-fields-remove-prefix) ✔ TF_ACC=1 go test -v -run='TestAccWorkItem_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItem_basic
=== PAUSE TestAccWorkItem_basic
=== RUN   TestAccWorkItem_titleUpdate
=== PAUSE TestAccWorkItem_titleUpdate
=== RUN   TestAccWorkItem_tagUpdate
=== PAUSE TestAccWorkItem_tagUpdate
=== RUN   TestAccWorkItem_parent
=== PAUSE TestAccWorkItem_parent
=== RUN   TestAccWorkItem_parentUpdate
=== PAUSE TestAccWorkItem_parentUpdate
=== RUN   TestAccWorkItem_parentDelete
=== PAUSE TestAccWorkItem_parentDelete
=== RUN   TestAccWorkItem_additionalFieldsJson
=== PAUSE TestAccWorkItem_additionalFieldsJson
=== RUN   TestAccWorkItem_description
=== PAUSE TestAccWorkItem_description
=== CONT  TestAccWorkItem_basic
=== CONT  TestAccWorkItem_parentUpdate
=== CONT  TestAccWorkItem_additionalFieldsJson
=== CONT  TestAccWorkItem_parentDelete
=== CONT  TestAccWorkItem_description
=== CONT  TestAccWorkItem_parent
=== CONT  TestAccWorkItem_titleUpdate
=== CONT  TestAccWorkItem_tagUpdate
--- PASS: TestAccWorkItem_basic (36.58s)
--- PASS: TestAccWorkItem_parent (38.51s)
--- PASS: TestAccWorkItem_additionalFieldsJson (39.14s)
--- PASS: TestAccWorkItem_parentDelete (40.75s)
--- PASS: TestAccWorkItem_titleUpdate (41.63s)
--- PASS: TestAccWorkItem_parentUpdate (41.74s)
--- PASS: TestAccWorkItem_description (42.87s)
--- PASS: TestAccWorkItem_tagUpdate (44.36s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	45.320s
```

The `utils` unit tests do not appear to be running in the GitHub workflow due to the build tags. Attaching results here:

```
➜ terraform-provider-azuredevops (feature/workitem-custom-fields-remove-prefix) ✔ go test -v ./azuredevops/internal/utils/suppress -tags=utils
=== RUN   TestCaseDifference
=== RUN   TestCaseDifference/empty
=== RUN   TestCaseDifference/empty_vs_text
=== RUN   TestCaseDifference/different_text
=== RUN   TestCaseDifference/same_text
=== RUN   TestCaseDifference/same_text_different_case
--- PASS: TestCaseDifference (0.00s)
    --- PASS: TestCaseDifference/empty (0.00s)
    --- PASS: TestCaseDifference/empty_vs_text (0.00s)
    --- PASS: TestCaseDifference/different_text (0.00s)
    --- PASS: TestCaseDifference/same_text (0.00s)
    --- PASS: TestCaseDifference/same_text_different_case (0.00s)
=== RUN   TestWhitespaceJsonDifference
=== RUN   TestWhitespaceJsonDifference/empty
=== RUN   TestWhitespaceJsonDifference/empty_vs_json_string
=== RUN   TestWhitespaceJsonDifference/different_json_string
=== RUN   TestWhitespaceJsonDifference/same_json_string
=== RUN   TestWhitespaceJsonDifference/same_json_string_whitespace_difference
=== RUN   TestWhitespaceJsonDifference/same_json_string_different_order
--- PASS: TestWhitespaceJsonDifference (0.00s)
    --- PASS: TestWhitespaceJsonDifference/empty (0.00s)
    --- PASS: TestWhitespaceJsonDifference/empty_vs_json_string (0.00s)
    --- PASS: TestWhitespaceJsonDifference/different_json_string (0.00s)
    --- PASS: TestWhitespaceJsonDifference/same_json_string (0.00s)
    --- PASS: TestWhitespaceJsonDifference/same_json_string_whitespace_difference (0.00s)
    --- PASS: TestWhitespaceJsonDifference/same_json_string_different_order (0.00s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/utils/suppress	(cached)
```

azuredevops/internal/service/workitemtracking/resource_workitem.go
```diff
@@ -255,13 +287,32 @@ func expandCustomFields(d *schema.ResourceData, operations []webapi.JsonPatchOpe
...
 			From:  nil,
-			Path:  converter.String("/fields/Custom." + customFieldName),
+			Path:  converter.String("/fields/" + customFieldsPrefix + customFieldName),
 			Value: customFieldValue,
 		})
 	}
 	return operations
 }
 
+func expandAdditionalFields(d *schema.ResourceData, operations []webapi.JsonPatchOperation) ([]webapi.JsonPatchOperation, error) {
```
magodo: Shall we also handle the case when the fields are removed, by comparing the state and plan?

azuredevops/internal/service/workitemtracking/resource_workitem.go
```diff
@@ -330,17 +381,39 @@ func expandTags(d *schema.ResourceData, operations []webapi.JsonPatchOperation,
 	return operations
 }
 
-func flattenFields(d *schema.ResourceData, m *map[string]interface{}) {
+func flattenFields(d *schema.ResourceData, m *map[string]interface{}) error {
```
magodo, line 458: My previous idea was to filter all "System." fields out and keep track of the remaining fields in the `additional_fields_json`, excep the `Microsoft.VSTS.Common.StateChangeDate`. This way, we shall be able to also import it without ignoring `additional_fields_json`. WDYT?

azuredevops/internal/utils/suppress/string.go
```diff
@@ -11,3 +13,24 @@ import (
 func CaseDifference(_, old, new string, _ *schema.ResourceData) bool {
 	return strings.EqualFold(old, new)
 }
+
+// WhitespaceJsonDifference reports whether old and new, interpreted as json strings,
+// are equal excluding "whitespace-only" drift.
+func WhitespaceJsonDifference(k, old, new string, d *schema.ResourceData) bool {
```
magodo: This is already defined in `structure.SuppressJsonDiff`.

magodo (changes requested):

azuredevops/internal/service/workitemtracking/resource_workitem.go
brittandeyoung: My concern with reading ALL fields other than `System.` is that i keep finding more fields defined that are dynamic in nature. If we miss one in our "blacklist" it will cause perminant Diff for projects using the field until it is added to the black list. At that point we may as well manually add fields for each field for stability. Here are some more example fields I found that are dynamic from the API list fields doc:

```json
[
  {
      "helpText": "Test context, provided automatically by test infrastructure",
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.TCM.SystemInfo",
      "name": "System Info",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.TCM.SystemInfo"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.StateChangeDate",
      "name": "State Change Date",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.StateChangeDate"
    },
        {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.StateChangeDate",
      "name": "State Change Date",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.StateChangeDate"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ActivatedDate",
      "name": "Activated Date",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ActivatedDate"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ActivatedBy",
      "name": "Activated By",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ActivatedBy"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ResolvedDate",
      "name": "Resolved Date",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ResolvedDate"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ResolvedBy",
      "name": "Resolved By",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ResolvedBy"
    },
    {
      "helpText": "The reason why the bug was resolved",
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ResolvedReason",
      "name": "Resolved Reason",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ResolvedReason"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ClosedDate",
      "name": "Closed Date",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ClosedDate"
    },
    {
      "alwaysRequired": false,
      "defaultValue": null,
      "referenceName": "Microsoft.VSTS.Common.ClosedBy",
      "name": "Closed By",
      "url": "https://dev.azure.com/fabrikam/_apis/wit/fields/Microsoft.VSTS.Common.ClosedBy"
    },
}
```

Ultimately it is the decision of the provider team if you would prefer using the field blacklist design or the limit fields to ones we define model. The black list model comes with the risk that we could miss a field and would require code changes to fix it, the limiting fields to ones defines causes import issues. 

Please let me know which one you want me to proceed with and I will update the logic.

azuredevops/internal/utils/suppress/string.go
brittandeyoung: I did not see this initially, updated. thanks

azuredevops/internal/service/workitemtracking/resource_workitem.go
brittandeyoung: Logic has been added for updates of the added fields.

brittandeyoung:
@magodo I have addressed two of your comments and responded to the third. 

I am struggling with a few things in the code.

1. Making the `description` field both `Optional` and `Computed` would prevent this from being a breaking change; however, it would prevent users from being able to set this field to a blank string and removing the field from config would simply leave the last value set (This could be an unexpected behavior for users). Do we use both `Optional` and `Computed` and simply document the behavior? Or would this better be a breaking change?

2. With the `additional_fields_json` I need some direction on which method to use, the current code will only read fields from the API when they exist in the config. This WILL cause an initial Diff on import, but not cause a breaking change. Alternatively, we go the "blacklist" method of fields we do not read, and read the remaining fields always. This will fix the import Diff issue, but introduce the same potential breaking change behavior described for the `description` field above.

3. I wanted to throw out there a potential solution to the breaking change concerns. Many providers will introduce resources specific to a sub type of a resource with improvements that would otherwise be a breaking change and then suggest users move to the new resource. This would prevent breaking existing resources, while providing the option for migrating and getting the new features. typically these look like `azruredevops_workitem_<type>` example: `azuredevops_workitem_issue` , `azuredevops_workitem_userstory`, etc.

Please provide direction and I will proceed as you direct. 

Currently the fields I added a listed as `Optional` only and would likely cause a breaking change just for the `description` field. 

If preferred I can also split the `additional_fields_json` and `description` into their own PRS to separate out the Breaking change and the non breaking change.

magodo:
@brittandeyoung Thanks for making the changes and throwing the concerns!

1. I'd prefer keeping the `description` Optional and Computed, even though the user won't be able to zero it then. We shall add a "TODO" comment to mention to remove the `Computed` in the major release, and document this behavior in the docs.
2. With the additional example fields you've listed, I feel blacklist might not be ideal since we don't have the full list. Instead, I think we can proceed with the "state"-based tracking method you've been implemented. The only question remained is during import there expects a diff - as there is no additional_fields in the state, while the config (i.e. plan) and the remote both have the same value - which will results into a couple of "Add" operations for some already existing fields. Will these API calls succeed? Or they'll fail with error like "field already exists"?
3. I'd like to avoid creating specific sub type resources if possible.

brittandeyoung:
@magodo I have update the PR with your latest guidance. Please let me know if there are still any additional changes needed. Here are the latest acceptance test results:

```
➜ terraform-provider-azuredevops (feature/workitem-custom-fields-remove-prefix) ✔ TF_ACC=1 go test -v -run='TestAccWorkItem_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItem_basic
=== PAUSE TestAccWorkItem_basic
=== RUN   TestAccWorkItem_titleUpdate
=== PAUSE TestAccWorkItem_titleUpdate
=== RUN   TestAccWorkItem_tagUpdate
=== PAUSE TestAccWorkItem_tagUpdate
=== RUN   TestAccWorkItem_parent
=== PAUSE TestAccWorkItem_parent
=== RUN   TestAccWorkItem_parentUpdate
=== PAUSE TestAccWorkItem_parentUpdate
=== RUN   TestAccWorkItem_parentDelete
=== PAUSE TestAccWorkItem_parentDelete
=== RUN   TestAccWorkItem_additionalFieldsJson
=== PAUSE TestAccWorkItem_additionalFieldsJson
=== RUN   TestAccWorkItem_description
=== PAUSE TestAccWorkItem_description
=== CONT  TestAccWorkItem_basic
=== CONT  TestAccWorkItem_parentUpdate
=== CONT  TestAccWorkItem_tagUpdate
=== CONT  TestAccWorkItem_parent
=== CONT  TestAccWorkItem_titleUpdate
=== CONT  TestAccWorkItem_additionalFieldsJson
=== CONT  TestAccWorkItem_description
=== CONT  TestAccWorkItem_parentDelete
--- PASS: TestAccWorkItem_basic (39.05s)
--- PASS: TestAccWorkItem_parent (40.91s)
--- PASS: TestAccWorkItem_additionalFieldsJson (40.96s)
--- PASS: TestAccWorkItem_titleUpdate (42.07s)
--- PASS: TestAccWorkItem_parentUpdate (43.26s)
--- PASS: TestAccWorkItem_parentDelete (43.46s)
--- PASS: TestAccWorkItem_tagUpdate (46.79s)
--- PASS: TestAccWorkItem_description (48.49s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	49.678s


````

File: website/docs/r/workitem.html.markdown [hunk 4]
```diff
@@ -125,7 +125,9 @@ The following arguments are supported:
 
 * `custom_fields` - (Optional, **Deprecated** use `additional_fields_json` argument instead) Specifies a list with Custom Fields for the Work Item.
 
-* `description` - (Optional) A description for the Work Item.
+!> **Warning** To prevent a breaking change, the current `description` argument behavior will not remove the field from the Azure DevOps API when it is removed from the config or set to a blank value. When removed or set to a blank value this field will be `Computed` and contain the value received from the Azure DevOps API.
```
magodo: This seems redundant?

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

website/docs/r/workitem.html.markdown
brittandeyoung: @magodo  I have attempted an update to the description to be less redundant.

magodo (approved):
LGTM! Thank you!

```shell
💤 TF_ACC=1 go test -v -run='TestAccWorkItem_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccWorkItem_basic
=== PAUSE TestAccWorkItem_basic
=== RUN   TestAccWorkItem_titleUpdate
=== PAUSE TestAccWorkItem_titleUpdate
=== RUN   TestAccWorkItem_tagUpdate
=== PAUSE TestAccWorkItem_tagUpdate
=== RUN   TestAccWorkItem_parent
=== PAUSE TestAccWorkItem_parent
=== RUN   TestAccWorkItem_parentUpdate
=== PAUSE TestAccWorkItem_parentUpdate
=== RUN   TestAccWorkItem_parentDelete
=== PAUSE TestAccWorkItem_parentDelete
=== RUN   TestAccWorkItem_additionalFieldsJson
=== PAUSE TestAccWorkItem_additionalFieldsJson
=== RUN   TestAccWorkItem_description
=== PAUSE TestAccWorkItem_description
=== CONT  TestAccWorkItem_basic
=== CONT  TestAccWorkItem_parentUpdate
=== CONT  TestAccWorkItem_tagUpdate
=== CONT  TestAccWorkItem_description
=== CONT  TestAccWorkItem_titleUpdate
=== CONT  TestAccWorkItem_parent
=== CONT  TestAccWorkItem_additionalFieldsJson
=== CONT  TestAccWorkItem_parentDelete
--- PASS: TestAccWorkItem_basic (34.80s)
--- PASS: TestAccWorkItem_parent (35.39s)
--- PASS: TestAccWorkItem_additionalFieldsJson (36.09s)
--- PASS: TestAccWorkItem_titleUpdate (36.33s)
--- PASS: TestAccWorkItem_parentDelete (36.59s)
--- PASS: TestAccWorkItem_parentUpdate (36.61s)
--- PASS: TestAccWorkItem_tagUpdate (37.45s)
--- PASS: TestAccWorkItem_description (38.41s)
PASS
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests	38.414s
```

brittandeyoung:
@magodo Thank you! I am new to contributing to this provider. Is there a standard release cycle?

PR #1543, opened 2026-04-20
Added service endpoint type data sources

aklakina:
## All Submissions:

## Description

Added data source to query registered service endpoint types. PR 1/2 of #1505 

- [ ] Yes
- [x] No

## Related Issue(s)

Fix #0000


azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -0,0 +1,227 @@
...
+	if serviceEndpointTypesCache == nil || len(*serviceEndpointTypesCache) == 0 {
+		return diag.FromErr(fmt.Errorf("no service endpoint types found"))
+	}
+
+	name := d.Get("name").(string)
+	authScheme := d.Get("authorization_scheme").(string)
+
+	if name == "" {
+		return diag.FromErr(fmt.Errorf("'name' must be specified"))
+	}
```
magodo: Instead of doing this, you can add a validation function for the `name` in the schema.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_types.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	results := make([]interface{}, 0)
+	for _, t := range *types {
+		serviceType := map[string]interface{}{}
+
+		if t.Name != nil {
+			serviceType["name"] = *t.Name
+			// Use name as id since there's no dedicated Id field
+			serviceType["id"] = *t.Name
```
magodo, line 114: Any reason to define the `id` here at the first place?

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -0,0 +1,227 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the service endpoint type",
+			},
+			"authorization_scheme": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The authorization scheme to retrieve parameters for",
+			},
+			"id": {
```
magodo: This shall not be declared as it conflicts with the built-in `id`.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// Extract parameters from InputDescriptors
+	parameters := make(map[string]string)
+	if foundType.InputDescriptors != nil {
+		for _, descriptor := range *foundType.InputDescriptors {
+			if descriptor.Id != nil {
+				defaultValue := ""
+				if descriptor.Values != nil && descriptor.Values.DefaultValue != nil {
+					defaultValue = *descriptor.Values.DefaultValue
```
magodo: Any reason to only set the default value as the value of the descriptor?

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for _, scheme := range *endpointType.AuthenticationSchemes {
+		if scheme.Scheme != nil && strings.EqualFold(*scheme.Scheme, authScheme) {
+			authParams := make(map[string]string)
+			if scheme.InputDescriptors != nil {
+				for _, descriptor := range *scheme.InputDescriptors {
+					if descriptor.Id != nil {
+						defaultValue := ""
+						if descriptor.Values != nil && descriptor.Values.DefaultValue != nil {
+							defaultValue = *descriptor.Values.DefaultValue
```
magodo: Same here.

magodo (changes requested):

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
aklakina: I have modified the implementation to also export possible values for a descriptor if the api provides this information.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_types.go
aklakina: Not really, Removed this.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
aklakina: Fixed.

magodo:
Hi @aklakina, did you forget to push the new commits here?

aklakina:
Hi, it very much looks like it. Let me push it :D

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -59,19 +54,61 @@ func DataServiceEndpointType() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The parameter name (descriptor ID)",
+						},
+						"default_value": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Optional:    true,
```
magodo: Since the parent level is Computed only, this shall be also computed only.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -59,19 +54,61 @@ func DataServiceEndpointType() *schema.Resource {
...
+							Description: "The default value for this parameter, if provided by the API",
+						},
+						"possible_values": {
+							Type:        schema.TypeList,
+							Computed:    true,
+							Optional:    true,
```
magodo: Same here.

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -59,19 +54,61 @@ func DataServiceEndpointType() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The authorization parameter name (descriptor ID)",
+						},
+						"default_value": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Optional:    true,
```
magodo: Same here

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
```diff
@@ -59,19 +54,61 @@ func DataServiceEndpointType() *schema.Resource {
...
+							Description: "The default value for this parameter, if provided by the API",
+						},
+						"possible_values": {
+							Type:        schema.TypeList,
+							Computed:    true,
+							Optional:    true,
```
magodo: Same here.

magodo (changes requested):

magodo:
Out of curiosity, how is the user supposed to use this data source in a real world use case?

azuredevops/internal/service/serviceendpoint/data_serviceendpoint_type.go
aklakina: Removed

aklakina:
Hi @magodo,

Sorry I didn't see your comment.

So basically this is supposed to be a complement feature for the 2nd pr of this group. One can query the data types and use a terraform_data resource block to print the content to stdout so they can check the properties a service connection need for example for authentication or for data blocks. It has no usecase other than that really.

magodo:
@aklakina Printing the content to stdout for reference of authoring another resource is not a typical data source use case. Instead, these information shall be in the document of the target resource.

aklakina:
@magodo I would agree with your comment there on generic cases. But this api queries a plugin system and the currently installed list of plugins. If you would want to have this in the documentation of the resource itself, you would either have to follow the so far established per type service endpoint resource (which is a bad practice for automation) or you would need to extend the documentation of the generic resource with every plugin there is.

This data source should be the documentation as noone can ensure that the plugins follow a specific naming convention or even that someone does not create a plugin which has the same type as another plugin but completely different structure.

aklakina:
Hi @magodo 

So do you think it makes sense to merge this?

magodo (approved):

PR #1565, opened 2026-06-05
fix: migrate data_users to ReadContext to enforce read timeout

ni-var:
## Description

The `data.azuredevops_users` data source was using the legacy `Read` callback instead of `ReadContext`. With the legacy callback, Terraform never cancels the context when the declared timeout fires — meaning the underlying HTTP call to the Azure DevOps Graph API can hang indefinitely, blocking the entire plan until the CI runner's hard wall-clock limit is reached.

This was observed in production: a `terraform plan` hung for over 1h20m on `data.azuredevops_users` lookups before being killed by the 1h30m TFC runner timeout. The same lookups completed in 7–8 seconds on a prior run, confirming the hang was due to the AzDO Graph API not responding rather than slow but progressing work.

**Root cause:** The Terraform Plugin SDK only wires timeout cancellation through `ReadContext` (and its siblings `CreateContext`, `UpdateContext`, `DeleteContext`). The legacy `Read` callback receives no `context.Context`, so the `Timeouts` block declared on the resource is effectively ignored. See:

- [Terraform Plugin SDK — Retries and Customizable Timeouts](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts): *"If the practitioner or the provider's timeout is reached, the context passed to `ReadContext` will be cancelled, but there is no mechanism for the legacy `Read` function."*
- [SDK v2 Upgrade Guide — Context-aware functions](https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#context-aware-functions)
- [SDK source — `resource.go`](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource.go): `ReadContext` receives a `ctx` derived from `d.Timeout(schema.TimeoutRead)`; legacy `Read` does not.
- [HashiCorp blog — Calling APIs without context in Terraform providers is risky](https://www.hashicorp.com/blog/calling-apis-without-context-in-terraform-providers-is-risky)

**Changes:**
- Migrate `DataUsers()` from `Read` to `ReadContext`
- Thread `context.Context` through `getUsersWithContinuationToken` and `addStorageKeyAsId` so in-flight HTTP requests are cancelled when the timeout fires
- Reduce the default read timeout from 30 minutes to 5 minutes — typical response time is under 10 seconds; 5 minutes is a safe ceiling that prevents plans from silently blocking for extended periods
- Update docs to reflect the new 5-minute default

- [ ] Yes
- [x] No

Users who explicitly configure `timeouts { read = "..." }` in their HCL are unaffected. The reduced default only impacts users relying on the implicit 30-minute default, and only in the (previously unbounded) hang scenario this fixes.

## Test Result

Existing unit and acceptance tests pass unchanged. The failure mode (AzDO Graph API returning no response) cannot be reliably reproduced in a test environment.

### DIFF
--- a/azuredevops/internal/service/graph/data_users.go
+++ b/azuredevops/internal/service/graph/data_users.go
@@ -1,6 +1,7 @@
 package graph
 
 import (
+	"context"
 	"crypto/sha1"
 	"encoding/base64"
 	"fmt"
@@ -9,6 +10,7 @@ import (
 	"time"
 
 	"github.com/ahmetb/go-linq"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 	"github.com/microsoft/azure-devops-go-api/azuredevops/v7/graph"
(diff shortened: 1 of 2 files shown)


magodo (approved):

ni-var:
@microsoft-github-policy-service agree company="NI"

PR #1570, opened 2026-06-23
`azuredevops_group_membership` - Fix 60 minute loop when mode="overwrite"

### LINKED ISSUE #1567: azuredevops_group_membership with mode = "overwrite" hangs ~60 min on create — StateChangeConf never reaches "Synched"
### Community Note

Terraform: 1.15.6
Provider: microsoft/azuredevops v1.15.1

### Affected Resource(s)

* azuredevops_group_membership

data "azuredevops_group" "pa" {
  project_id = azuredevops_project.example.id
  name       = "Project Administrators"
}

resource "azuredevops_group_membership" "pa" {
  group   = data.azuredevops_group.pa.descriptor
  mode    = "overwrite"
  members = [
    # a mix of an Entra group (aadgp.), a project group (vssgp.) and a
    # service principal / managed identity (aadsp.) — all valid, all already
    # the exact desired members
  ]

  timeouts {
    create = "5m"   # ignored — see below
  }
}

### Expected Behavior

After adding the members, the resource should detect that the group's actual membership matches the desired set and complete quickly.

### Actual Behavior

The members are added (verified via the Graph REST API — the group ends up with exactly the desired members), but the resource stays in Still creating... for ~60 minutes and then the operation is killed by the surrounding CI timeout. The timeouts.create value is ignored.

### Possible Root Cause
(from source, v1.15.1)

In azuredevops/internal/service/graph/resource_group_membership.go, resourceGroupMembershipCreate:

The wait is hardcoded and runs on the provider context, ignoring the resource timeouts:

stateConf := &retry.StateChangeConf{
    ...
    Timeout: 60 * time.Minute,
}
stateConf.WaitForStateContext(clients.Ctx) // not the resource create-timeout ctx
The "Synched" condition treats membersToAdd backwards — it only converges when the just-added members are absent from the live membership:

if (membersToAdd == nil || actualMembershipsSet.Intersection(membersToAdd).Len() <= 0) &&
   (membersToRemove == nil || actualMembershipsSet.Intersection(membersToRemove).Len() <= 0) {
    state = "Synched"
}
Since the members were successfully added, actual ∩ membersToAdd equals membersToAdd (non-zero), so state stays "Waiting" until the 60-minute timeout. The Update path has the same expression. This looks inverted: convergence for membersToAdd should require those members to be present (e.g. actual ∩ membersToAdd == membersToAdd), not absent.
This only manifests on the create path; Update short-circuits via if !d.HasChange("members"), so existing memberships that don't change are unaffected — which is why add-mode and already-created memberships don't visibly hang.

### Steps to Reproduce

terraform apply a fresh azuredevops_group_membership with mode = "overwrite" and a non-empty members list against a newly created project.
Observe the members get added in ADO almost immediately.
Observe the resource hangs ~60 minutes, then times out; timeouts.create is ignored.

### References

Possibly related: 
#291 (overwrite not applying on first run), 
#931 (unsafe behavior with pre-existing members). 

The create-time fix #1501 (in 1.14.0) does not resolve this.

v-kenntan:
## All Submissions:

## Description

Fixes a bug in  azuredevops_group_membership  where, with mode = "overwrite" and a non-empty target group at create time, the members are added successfully but the post-create sync hangs for the full 60-minute timeout.

Also fixed timeout to honor the timeout block instead of the 60 minutes hardcode.

### Root Cause

`membersToAdd` was initially set to `d.Get("members").(*schema.Set)`, and while the values inside are correct, `actualMembershipsSet` is populated using `schema.NewSet(...)`. This means that using the `Difference` and `Intersection` functions between `membersToAdd` and `actualMembershipsSet` won't work as intended. In this case, using `actualMembershipsSet.Difference(membersToAdd)` will always return all the values inside `actualMembershipsSet`, which then lead to `membersToRemove` always containing the existing members inside the group. This also means `actualMembershipsSet.Intersection(membersToAdd)` will always be empty, hence the initial `if (membersToAdd == nil || actualMembershipsSet.Intersection(membersToAdd).Len() <= 0)` check.

This has been resolved by also populating `membersToAdd` using `schema.NewSet(...)` and fixing the logic inside the post create sync function.

- [ ] Yes
- [x] No

## Test Result
<pre>
=== RUN   TestAccGroupMembership_overwrite
=== PAUSE TestAccGroupMembership_overwrite
=== CONT  TestAccGroupMembership_overwrite
--- PASS: TestAccGroupMembership_overwrite (142.15s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        142.159s
</pre>

## Related Issue(s)

Fix #1567

### DIFF
--- a/azuredevops/internal/service/graph/resource_group_membership.go
+++ b/azuredevops/internal/service/graph/resource_group_membership.go
@@ -61,7 +61,7 @@ func resourceGroupMembershipCreate(d *schema.ResourceData, m interface{}) error 
 	clients := m.(*client.AggregatedClient)
 	group := d.Get("group").(string)
 	mode := d.Get("mode").(string)
-	membersToAdd := d.Get("members").(*schema.Set)
+	membersToAdd := schema.NewSet(schema.HashString, d.Get("members").(*schema.Set).List())
 	var membersToRemove *schema.Set = nil
 
 	if strings.EqualFold("overwrite", mode) {
@@ -93,14 +93,14 @@ func resourceGroupMembershipCreate(d *schema.ResourceData, m interface{}) error 
 				return nil, "", fmt.Errorf("Reading group memberships: %+v", err)
 			}
 			actualMembershipsSet := getGroupMembershipSet(actualMemberships)
-			if (membersToAdd == nil || actualMembershipsSet.Intersection(membersToAdd).Len() <= 0) &&
-				(membersToRemove == nil || actualMembershipsSet.Intersection(membersToRemove).Len() <= 0) {
+			if (membersToAdd == nil || actualMembershipsSet.Intersection(membersToAdd).Len() == membersToAdd.Len()) &&
+				(membersToRemove == nil || actualMembershipsSet.Intersection(membersToRemove).Len() == 0) {
 				state = "Synched"
 			}
 
 			return state, state, nil
 		},
-		Timeout:                   60 * time.Minute,
+		Timeout:                   d.Timeout(schema.TimeoutCreate),
 		MinTimeout:                5 * time.Second,
 		Delay:                     5 * time.Second,
 		ContinuousTargetOccurence: 2,
@@ -151,9 +151,9 @@ func resourceGroupMembershipUpdate(d *schema.ResourceData, m interface{}) error 
 	group := d.Get("group").(string)
 	oldData, newData := d.GetChange("members")
 	// members that need to be added will be missing from the old data, but present in the new data
-	membersToAdd := newData.(*schema.Set).Difference(oldData.(*schema.Set))
+	membersToAdd := schema.NewSet(schema.HashString, newData.(*schema.Set).Difference(oldData.(*schema.Set)).List())
 	// members that need to be removed will be missing from the new data, but present in the old data
-	membersToRemove := oldData.(*schema.Set).Difference(newData.(*schema.Set))
+	membersToRemove := schema.NewSet(schema.HashString, oldData.(*schema.Set).Difference(newData.(*schema.Set)).List())
 
 	err := applyMembershipUpdate(m.(*client.AggregatedClient),
 		expandGroupMembers(group, membersToAdd),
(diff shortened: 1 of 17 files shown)


azuredevops/internal/service/graph/resource_group_membership.go
```diff
@@ -151,9 +151,9 @@ func resourceGroupMembershipUpdate(d *schema.ResourceData, m interface{}) error
 	group := d.Get("group").(string)
 	oldData, newData := d.GetChange("members")
 	// members that need to be added will be missing from the old data, but present in the new data
-	membersToAdd := newData.(*schema.Set).Difference(oldData.(*schema.Set))
+	membersToAdd := schema.NewSet(schema.HashString, newData.(*schema.Set).Difference(oldData.(*schema.Set)).List())
 	// members that need to be removed will be missing from the new data, but present in the old data
-	membersToRemove := oldData.(*schema.Set).Difference(newData.(*schema.Set))
+	membersToRemove := schema.NewSet(schema.HashString, oldData.(*schema.Set).Difference(newData.(*schema.Set)).List())
```
magodo: What is this for?

magodo (changes requested):

azuredevops/internal/service/graph/resource_group_membership.go
v-kenntan: It's for the same case as in the Create function. In the post create sync function, it uses `getGroupMembershipSet`, which populates the group membership set using `schema.NewSet(...)`. This then causes the `Intersection` function to not find any matches. That's why the if statement used to be `actualMembershipsSet.Intersection(membersToAdd).Len() <=0`, when it should be `actualMembershipsSet.Intersection(membersToAdd).Len() == membersToAdd.Len()`.

magodo: No, this highlighted part seems to be doing the same thing as the original version.

v-kenntan: For the `Difference` function in the highlighted part, it is working fine since both `oldData` and `newData` were populated using this: `oldData, newData := d.GetChange("members")`. However, `membersToAdd` and `membersToRemove` would then need to be converted using `schema.NewSet` for the `Intersection` function in the post create sync function.

magodo: I mean this change seems unnecessary, the original version is cleaner.

v-kenntan:
Hi, just changed to use the `go-set` library.

<h4>Acceptance Tests</h4>
<pre>
=== RUN   TestAccGroupMembership_overwrite
=== PAUSE TestAccGroupMembership_overwrite
=== CONT  TestAccGroupMembership_overwrite
--- PASS: TestAccGroupMembership_overwrite (137.13s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        137.143s
</pre>

magodo (approved):

Issue #1582, opened 2026-07-15
Creation of teams fails


rcomanne:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform version: 1.15.8-1
Provider version: v1.15.1

### Affected Resource(s)

* `azuredevops_team`

### Terraform Configuration Files

We use a module for managing teams, this was working fine but now whenever we add a team, it seems to hang until it stops trying after 30mins.  
The administrators/members passed are always descriptors, either from an existing team or from an EntraID group which is already known to Azure DevOps.

```hcl
data "azuredevops_group" "administrators" {
  for_each = var.administrator_groups

  project_id = var.project_id
  name       = each.value
}

data "azuredevops_group" "aad_administrators" {
  for_each = var.aad_administrator_groups

  name = each.value
}

data "azuredevops_group" "aad_members" {
  for_each = var.aad_member_groups

  name = each.value
}

locals {
  combined_members = concat(
    [for group in data.azuredevops_group.aad_members : group.descriptor],
  )

  combined_administrators = concat(
    [for group in data.azuredevops_group.administrators : group.descriptor],
    [for group in data.azuredevops_group.aad_administrators : group.descriptor]
  )

  administrators = length(local.combined_administrators) > 0 ? local.combined_administrators : local.combined_members
}

resource "azuredevops_team" "team" {
  project_id  = var.project_id
  name        = var.team_name
  description = "[Terraform Managed] ${var.team_description}"

  # Not needed for initial setup. Keeping it clear to read what's changing. 
  lifecycle {
    ignore_changes = [
      description
    ]
  }

  administrators = local.administrators
  members        = local.combined_members
}

```

### Expected Behavior

Run completed without issues, with the teams created and stored correctly in state.

### Actual Behavior

Terraform apply fails, after trying to read back for 30 minutes even though the teams exist in Azure DevOps. Nothing is stored in state either.

```
╷
[...]

Issue #1464, opened 2025-12-11
Version 1.12.0 breaks azuredevops_team  .. it wil never reach "Synced" state.


rob-simac:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform (and Azure DevOps Provider) Version

Terraform 1.12.0
Provider 1.12.0

### Affected Resource(s)

azuredevops_team

### Debug Output

```
2025-12-11T13:19:18.9686320Z 2025-12-11T13:19:18.967Z [TRACE] dag/walk: vertex "module.generic_devops_projects (close)" is waiting for "module.generic_devops_projects.module.teams (close)"
2025-12-11T13:19:18.9755267Z 2025-12-11T13:19:18.975Z [INFO]  provider.terraform-provider-azuredevops_v1.12.0: 2025/12/11 13:19:18 [TRACE] Waiting 10s before next try: timestamp=2025-12-11T13:19:18.975Z
2025-12-11T13:19:18.9806526Z 2025-12-11T13:19:18.980Z [TRACE] dag/walk: vertex "root" is waiting for "module.generic_devops_projects[\"XXXXX\"].module.teams[\"XXXX\"].output.id"
2025-12-11T13:19:18.9941955Z 2025-12-11T13:19:18.993Z [INFO]  provider.terraform-provider-azuredevops_v1.12.0: 2025/12/11 13:19:18 [TRACE] Waiting 10s before next try: timestamp=2025-12-11T13:19:18.993Z
2025-12-11T13:19:18.9943292Z 2025-12-11T13:19:18.994Z [INFO]  provider.terraform-provider-azuredevops_v1.12.0: 2025/12/11 13:19:18 [TRACE] Waiting 10s before next try: timestamp=2025-12-11T13:19:18.994Z
2025-12-11T13:19:19.0703317Z 2025-12-11T13:19:19.069Z [INFO]  provider.terraform-provider-azuredevops_v1.12.0: 2025/12/11 13:19:19 [TRACE] Waiting 10s before next try: timestamp=2025-12-11T13:19:19.069Z
[...]

PR #1599, opened 2026-09-02
`azuredevops_team` - Fix create/update never reaching `Synched` when …

itmagix:
## All Submissions:

## Description

While migrating a few hundred teams to a new organization I ran into #1582: team creation hangs for 30 minutes and then fails with `timeout while waiting for state to become 'Synched' (last state: 'Waiting')`, even though the team and its members were created fine in Azure DevOps. Nothing lands in state, so the next run or an import picks the teams up immediately.

The convergence loop in `waitForTeamStateChange()` has five conditions that all have to hold, and several of them can never become true under normal service behavior:

- name and description are compared with exact string equality. The service trims surrounding whitespace when it stores them, so the comparison can never pass. A missing value in the response also dereferenced a nil pointer.
- members and administrators are compared by count only. The service can seed its own ACEs on the team identity token, report extra or duplicate entries, or silently drop a membership, so the counts never match even when everything the configuration asked for is there.
- the dashboard check reads `if dashboards == nil && len(*dashboards) == 0`, which dereferences a nil pointer. It could only ever crash or pass.
- the 30 minute wait timeout was hardcoded and ignored the resource `timeouts` configuration.

Because the loop does not log which condition is stuck, every one of these quirks produces the same silent 30 minute timeout. The same symptom came up in #1464, where the actual cause turned out to be an unrelated regression in `GetAccessControlList()` (fixed by the revert in #1466).

This PR changes the loop to:

- compare trimmed name and description values, and accept a missing value in the response instead of panicking or waiting forever
- require that all configured members and administrators are present (set intersection, the same semantics the standalone `azuredevops_team_members` resource already uses) and log a warning when the counts differ
- treat a nil or empty dashboard list as a warning instead of panicking
- take the wait timeout from the resource `timeouts` configuration instead of the hardcoded 30 minutes
- skip identities without a subject descriptor in `getSubjectDescriptors()`

No documentation changes are needed. There is no schema change, the documented `timeouts` block simply works now.

- [ ] Yes
- [x] No

Config compatible. Two behavior changes worth calling out: the default create/update wait is now the documented 10 minute resource default instead of a hardcoded 30 minutes, and the loop converges on set intersection instead of exact counts, logging a warning when the team state differs from the configuration.

## Test Result

```
--- PASS: TestTeam_WaitForStateChange_Synced (15.01s)
--- PASS: TestTeam_WaitForStateChange_Converges_WhenServiceNormalizesNameAndDescription (15.01s)
--- PASS: TestTeam_WaitForStateChange_Converges_WhenResponseOmitsDescription (15.01s)
--- PASS: TestTeam_WaitForStateChange_Converges_WhenACLContainsExtraAdministratorACE (15.00s)
--- PASS: TestTeam_WaitForStateChange_Converges_WhenTeamHasExtraMembers (15.01s)
--- PASS: TestTeam_WaitForStateChange_DoesNotPanic_WhenDashboardsListIsNil (15.01s)
--- PASS: TestTeam_WaitForStateChange_TimesOut_WhenConfiguredMemberNeverAppears (12.01s)
ok  	github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/service/core	102.071s
```

`go build`, `go vet` and the full core package test suite pass. The patched provider has also been verified in a production pipeline while migrating teams between organizations: teams and members are created and the loop converges in seconds instead of hanging for 30 minutes.

## Related Issue(s)

Addresses #1582. The same symptom came up in #1464, caused by an unrelated regression and fixed by the revert in PR #1466.

## Other information

The triggers for the original hang are all realistic service side behaviors: normalized name and description (confirmed in the #1582 thread), extra ACEs on the team identity namespace, and memberships the identity service does not persist. Each one used to produce the same 30 minute timeout with no indication of which condition was stuck. The loop now logs the drifted condition, which should make reports like #1582 much easier to diagnose.

### DIFF
--- a/azuredevops/internal/service/core/resource_team.go
+++ b/azuredevops/internal/service/core/resource_team.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/ahmetb/go-linq"
@@ -187,7 +188,7 @@ func resourceTeamCreate(d *schema.ResourceData, m interface{}) error {
 		}
 	}
 
-	if err = waitForTeamStateChange(d, clients, projectID, teamID, teamData.Name, teamData.Description, memberSet, administratorSet, areaSet); err != nil {
+	if err = waitForTeamStateChange(d, clients, projectID, teamID, teamData.Name, teamData.Description, memberSet, administratorSet, areaSet, d.Timeout(schema.TimeoutCreate)); err != nil {
 		return err
 	}
 
@@ -350,7 +351,7 @@ func resourceTeamUpdate(d *schema.ResourceData, m interface{}) error {
 		}
 	}
 
-	if err := waitForTeamStateChange(d, clients, projectID, teamID, newTeamName, newDescription, memberSet, administratorSet, areaSet); err != nil {
+	if err := waitForTeamStateChange(d, clients, projectID, teamID, newTeamName, newDescription, memberSet, administratorSet, areaSet, d.Timeout(schema.TimeoutUpdate)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


itmagix:

rcomanne:
Can someone have a look at this? This will fix a long running issue for us where creating teams in one shot does not work.

PR #1598, opened 2026-09-02
`azuredevops_serviceendpoint_*` resources - Fix fail to destroy error due to consistency issues

v-kenntan:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## Description

Currently, in `azuredevops_serviceendpoint_*` resources, when the user tries to delete the resource, an error occurs if `azurerm_federated_identity_credential` is deleted, but due to consistency issues, the API returns it as not deleted, which then causes a `Delete service endpoint error Cannot delete this service connection while federated credentials for app <id of app> exist in Entra tenant <id of WIF>` error. This fix adds a retry to prevent it from erroring out due to consistency issues.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Test Result

<pre>
=== RUN   TestAccServiceEndpointAzureCR_dataSource
=== PAUSE TestAccServiceEndpointAzureCR_dataSource
=== RUN   TestAccServiceEndpointAzureCR_spn_basic
=== PAUSE TestAccServiceEndpointAzureCR_spn_basic
=== RUN   TestAccServiceEndpointAzureCR_spn_update
=== PAUSE TestAccServiceEndpointAzureCR_spn_update
=== RUN   TestAccServiceEndpointAzureCR_workLoadIdentity_basic
=== PAUSE TestAccServiceEndpointAzureCR_workLoadIdentity_basic
=== RUN   TestAccServiceEndpointAzureCR_workLoadIdentity_update
=== PAUSE TestAccServiceEndpointAzureCR_workLoadIdentity_update
=== CONT  TestAccServiceEndpointAzureCR_dataSource
=== CONT  TestAccServiceEndpointAzureCR_workLoadIdentity_basic
=== CONT  TestAccServiceEndpointAzureCR_workLoadIdentity_update
=== CONT  TestAccServiceEndpointAzureCR_spn_basic
=== CONT  TestAccServiceEndpointAzureCR_spn_update
--- PASS: TestAccServiceEndpointAzureCR_dataSource (54.11s)
--- PASS: TestAccServiceEndpointAzureCR_spn_basic (54.27s)
--- PASS: TestAccServiceEndpointAzureCR_spn_update (56.33s)
[...]

PR #1523, opened 2026-03-12
`azuredevops_group(s)`: Added filter to only filter out project-scoped groups when `project_id` is specified

v-kenntan:
## All Submissions:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] Have you checked to ensure there aren't other open PRs for the same update/change?

## Description

Added filter to only filter out project-scoped groups when project_id is specified for azuredevops_group and azuredevops_groups. Currently when a project id is specified, all groups are included regardless of scope.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Test Result
<pre>
TF_ACC=1 go test -v -run='TestAccGroupsDataSource_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccGroupsDataSource_Read_Project
=== PAUSE TestAccGroupsDataSource_Read_Project
=== RUN   TestAccGroupsDataSource_Read_NoProject
=== PAUSE TestAccGroupsDataSource_Read_NoProject
=== RUN   TestAccGroupsDataSource_ProjectID_FiltersOutCollectionGroups
--- PASS: TestAccGroupsDataSource_ProjectID_FiltersOutCollectionGroups (28.26s)
=== CONT  TestAccGroupsDataSource_Read_Project
=== CONT  TestAccGroupsDataSource_Read_NoProject
--- PASS: TestAccGroupsDataSource_Read_NoProject (12.50s)
--- PASS: TestAccGroupsDataSource_Read_Project (29.05s)
PASS
ok      github.com/microsoft/terraform-provider-azuredevops/azuredevops/internal/acceptancetests        57.319s
</pre>

<pre>
TF_ACC=1 go test -v -run='TestAccGroupDataSource_' ./azuredevops/internal/acceptancetests
=== RUN   TestAccGroupDataSource_Read_HappyPath
=== PAUSE TestAccGroupDataSource_Read_HappyPath
=== RUN   TestAccGroupDataSource_Read_ProjectCollectionAdministrators
=== PAUSE TestAccGroupDataSource_Read_ProjectCollectionAdministrators
=== RUN   TestAccGroupDataSource_ReadersResolvesWithProjectID
--- PASS: TestAccGroupDataSource_ReadersResolvesWithProjectID (25.28s)
=== CONT  TestAccGroupDataSource_Read_HappyPath
[...]

PR #1604, opened 2026-09-16
Run the full unit test suite in CI

### LINKED ISSUE #1439: Full unit test suite isn't running on CI
If you take a look at the unit tests running in CI, it appears that they aren't all run.

See this run as an example:

https://github.com/microsoft/terraform-provider-azuredevops/actions/runs/17143816385/job/48652393554?pr=1417

Firstly, is this intentional?

If it's not, perhaps we could update the `make test` command to provide the `all` tag that seems to be missing?

The other issue is that when it run all the tests locally, a few of them fail. I haven't dived into this yet to see if it's a legitimate failure or an issue with my environment.

Anyone else seeing this?

### Community Note

N/A

### Affected Resource(s)

Unit tests are not being run - an example is: `data_projects_test.go`

N/A

### Debug Output

See CI test run here:

https://github.com/microsoft/terraform-provider-azuredevops/actions/runs/17143816385/job/48652393554?pr=1417

### Panic Output

N/A

### Expected Behavior

All tests apart from acceptance should run.

### Actual Behavior

A subset of all tests are run.

### Steps to Reproduce

1. Clone and create a development environment
2. Run `make test`

Or run CI on the latest main.

### Important Factoids

N/A

### References

N/A

sary8:
## All Submissions:

On the tests checkbox: nothing new to cover here, this turns 499 existing unit tests back
on and repairs them. The only new test code is six `..._Update_DoesNotSwallowError`
helpers, and those are missing helpers rather than new coverage — see below.

## Description

`make test` is `go test -v ./...` with no tag, so every `_test.go` behind
`//go:build all || ...` gets skipped. CI compiles 35 of the 536 unit tests in the repo.

This is the second half of what @magodo described in #1439: repair the UTs that rotted
while nobody was running them, then drop the tags the way #1436 did for the acceptance
tests. The one acceptance file #1436 missed (`resource_servicehook_webhook_tfs_test.go`)
is in here too. 534 unit tests pass afterwards — 536 minus two that turned out to be
exact duplicates.

### Reviewing this

113 files, but most of it is mechanical. The tag removal commit deletes constraint lines
and nothing else; this prints nothing:

```bash
git show <tag-removal-sha> -- '*_test.go' \
  | grep -E '^[-+]' | grep -vE '^(\+\+\+|---)' \
  | grep -vE '^[-+]\s*//(go:build|\s*\+build)' | grep -vE '^[-+]$'
```

What's worth reading is the two test-repair commits, `.golangci.yml`, and three product
files: one line in `build/resource_build_definition.go`, plus the sleep tunables in
`serviceendpoint/commons.go` and `graph/resource_group_membership.go`.

Every commit is green on its own, but that doesn't mean much before the tags come off —
nothing in those files is compiled yet. The verification that counts is at the tip.

### What was broken

Two packages didn't compile:

* `graph/data_users_test.go` still calls `dataUsersRead` and the two-result `flattenUser`;
  both changed in #1565 when the data source moved to `ReadContext`.
* 13 call sites in `serviceendpoint` still take an `error` back from
  `expandServiceEndpoint*`.

55 tests failed:

* 4 look for a leading space in an error string that was removed when the wrapping was
  cleaned up (`identity`, `graph`).
* 2 in `data_group(s)_test.go` build groups with no `domain`, so the project-scoped filter
  from #1523 drops all of them.
* 32 delete expectations allow a single call; `deleteServiceEndpoint` retries three times
  since #1598.
* 2 are the `triggers` regression below.

16 tested the wrong thing. Every `Test*_Update_DoesNotSwallowError*` in `argocd`,
`artifactory`, `jenkins`, `maven`, `nexus` and the four `jfrog_*_v2` files calls its own
file's *delete* helper, so the update path has never run and the delete scenario ran
twice. Six of those files have no update helper at all, so I wrote them.

2 passed by accident — `go test -shuffle=on ./...` fails on main's ordering:

* `sortBuildDefinition` rewrites `branchFilters`/`pathFilters` inside the package-level
  trigger maps, which `testBuildDefinition*WithCITrigger` share.
* `dataUsersReadContext` filters with `linq ... ToSlice(&newUsers)`, and go-linq writes
  into the backing array it is handed, so `&usrList1` gets clobbered.
  `TestDataSourceUser_Read_TestFilterByOriginId` expected three `GetStorageKey` calls from
  a fixture holding two matching users; the third came from a duplicate the previous test
  left behind.

Plus dead fixtures, an `append` whose result is thrown away, and JFrog/Maven project ids
that aliased the Artifactory ones instead of their own — all of it surfaced by
`unused`/`staticcheck` once these files get linted.

### The product changes

`expandBuildDefinition` declares `var buildTriggers []any`, so a definition with no
trigger goes out as `"triggers": null`. `omitempty` doesn't save it: the field is a
`*[]interface{}` and the pointer is never nil. Before #1396 the line was
`buildTriggers := append(ciTriggers, ...)`, and `expandBuildDefinitionTriggerList` returns
`make([]interface{}, 0, len(d))`, so the provider had always sent `[]`. The round-trip
test asserts `[]` and is exactly the test that couldn't run.

I put the old payload back rather than relax the test, but I have no Azure DevOps org to
check whether the service treats the two forms differently — the case I'd worry about is
an update that removes the last trigger. For what it's worth, nobody has reported a
trigger removal failing in the fifteen months since #1396, so either the service reads
both the same way or the case is rare. That's commit 1 on its own, so if you'd rather keep
`null` and change the fixture, I'll drop it.

The other change is four `const` durations becoming package-level `var`s that `TestMain`
zeroes: `stateChangeDelay` and `deleteRetryInterval` in `serviceendpoint/commons.go`,
`membershipSyncDelay` and `membershipSyncMinTimeout` in `graph/resource_group_membership.go`.
Production values don't move. Without it the suite is unusable: 25 delete-error tests pay
10s of retry back-off each, `TestServiceEndpointAzureRM_CreateWithValidate_DoesNotSwallowError`
pays the 10s `StateChangeConf.Delay` once per fixture for 70s, and
`TestGroupMembership_Destroy_DoesNotSwallowErrors` another 10s. `MinTimeout` is only
touched in `graph`, where `ContinuousTargetOccurence` forces a second poll. Tell me if you
want a different seam and I'll rework it.

### .golangci.yml

These 99 files get linted for the first time, which is 222 findings: 148 `SA1019` and
74 `errcheck`.

The 74 `errcheck` are fixed rather than excluded. 58 are `flatten*`/`expand*` helpers
called to seed a `*schema.ResourceData`, and every one of those call sites has a
`*testing.T` in scope, so they assert now. Two of them were worth having:
`TestGitRepo_Read_Uses{Id,Name}IfIdNotSet` stub `GetRepository` to fail and then dropped
the error `resourceGitRepositoryRead` hands back. The other 16 come from `check-blank` —
`uuid.NewUUID()` in files that already use `uuid.New()`, and so on.

That leaves one exclusion, for `SA1019`: a resource that registers the non-context
callbacks can only be driven through the deprecated `schema.Resource` fields. The rule
matches that specific message, so any other `SA1019` in a test still reports.

### Deleted tests

`TestServiceEndpointArgoCD_Update_DoesNotSwallowError` and
`TestServiceEndpointNexus_Update_DoesNotSwallowError`. Once the mis-wired `_Token` and
`_Password` variants in those files point at the update helper, these two call it with the
same fixture and the same arguments.

### Docs and tooling

`docs/testing.md` still requires a build tag on every `_test.go` file, which would have
sent the next contributor straight back into this. That section and the tag-based subset
examples are rewritten around `go test -run` (and the `TEST_ACC` typo a few lines down is
now `TF_ACC`). Same for `-tags`/`TESTTAGS` in `scripts/unittest.sh`, `scripts/acctest.sh`,
`GNUmakefile` and `README.md`, and for the `.vscode` settings and the `tf-azdo-test`
snippet — the gopls workaround #1436 wanted rid of. It's a separate commit if you'd rather
not have it here.

`acctest.sh`, `unittest.sh` and `make testacc` now **fail** when handed the old tag
argument rather than accepting and ignoring it. That matters most for the acceptance
tests, where the ignored argument means creating and destroying real resources: on main
`./scripts/acctest.sh resource_project` already selects 600 of the 607 acceptance tests,
because #1436 untagged them. So this is a footgun that predates the PR, but it would have
been made complete by it.

No CHANGELOG entry: that's updated in its own release PR, and #1436 didn't touch it. For
whenever the next one is written:

```
- `azuredevops_build_definition` - Send an empty `triggers` list instead of `null` when no trigger is configured (#XXXX)
```

### Not fixed here

`getGroupsForDescriptor` in `graph/data_group.go` dereferences `*grp.Domain` one line
before it checks `grp.Domain != nil`, so `azuredevops_group(s)` with `project_id` set will
panic on a group with no domain. The branch just above guards correctly. It has nothing to
do with #1439 so I left it alone; I can send it separately.

- [ ] Yes
- [x] No

Nothing changes for provider users beyond `azuredevops_build_definition` going back to
sending `"triggers": []`, which is what it did before #1396.

For contributors it is a break, and a deliberate one: `./scripts/acctest.sh <tag>`,
`./scripts/unittest.sh <tag>` and `make testacc TESTTAGS=<tag>` now exit with an error
instead of running.

## Test Result

windows/amd64, go 1.25.0, golangci-lint v2.12.0 (the version `golint.yml` pins).

```
main          go test ./...    35 unit tests   ~6s
this branch   go test ./...   534 unit tests   ~8s, green

./scripts/gofmtcheck.sh
go vet ./...
golangci-lint run ./...          0 issues
go mod tidy && go mod vendor     no diff
./scripts/terrafmt.sh
misspell -error -source=text website/
```

`go test -shuffle=on ./...` eight times, green each time. That's what surfaced the two
order-dependent tests above.

Nothing in the 499 newly enabled tests uses `t.Parallel()`, `time.Sleep`, `time.Now()`,
`os.Getenv`, `filepath` or `httptest` — they're all gomock, in memory.

No acceptance tests: I don't have an org to run them against. I'm glad to run them for any
resource you think is affected if you can point me at a setup, otherwise this will need
someone with access.

## Related Issue(s)

Fix #1439

#1436 did the same for the acceptance tests. The regressions repaired here came from
#1396, #1523, #1565 and #1598.

## Other information

Once this lands, PRs that touch a resource will start hitting unit tests that have been
asleep for a while. That's the point of #1439, but it's worth knowing.


sary8:

